PR #9, opened 2017-08-14
more launch options

fagoatse:


mastercoms/cfg/autoexec.cfg
```diff
@@ -2,8 +2,9 @@
 // Version: v3.4.0 | August 13 2017
 
 // Launch Options:
-// -novid -nojoy -noff -nohltv -high -nouserclip -softparticlesdefaultoff -reuse -usetcp -primarysound -snoforceformat -disable_d3d9_hacks
+// -console -novid -nojoy -noff -nohltv -high -nouserclip -softparticlesdefaultoff -reuse -usetcp -primarysound -snoforceformat -disable_d3d9_hacks
```
mastercoms: Is this launch option really necessary? Why not just have con_enable 1 in the autoexec?

fagoatse: -console saves you a single ~ press since afaik con_enable doesn't display it automatically. I find it useful to see that everything has loaded correctly when I start the game. It might be my OCD tho.

mastercoms: Then add it to the extra launch options instead of the defaults.

fagoatse: fair enough

PR #12, opened 2017-10-21
Skip some custom file downloads by default

fagoatse:
Let's make it default so that we don't clog our downloads folders with custom sounds, particles, models and other unwanted files.
"nosounds" is also an option here so it's up to you :)


mastercoms:
I don't think it should be default because the expected behavior for users would be to have that custom experience applied to them when they visit a certain server which provides these files. For example, VSH servers, Jailbreak, etc. These people might be confused when their server doesn't function properly.

fagoatse:
Fair enough. So perhaps shall we move it to comp/maxfps presets?

mastercoms:
I think it would be good there, especially since some comp servers are outdated and give you voicelines in old file locations, wasting download time and storage space.

PR #13, opened 2017-10-21
add cl_downloadfilter mapsonly to comp/maxperf presets

fagoatse:


mastercoms:
Could you add documentation like

`// Only downloads maps from server custom files`

PR #33, opened 2018-02-23
rate module based off of upload speed

JarateKing:
Based off of my jarconfig's implementation.

Doesn't automatically go with 80% lower than the user sets like through the old method, because ~8.4mbps being the max but also not when setting it would be confusing. Can be changed without much effort though.

Placed the module at the bottom of the net settings section because it's a giant block of aliases that wouldn't fit too well in the middle.


mastercoms:
Thank you for your contribution, but I've already created rate modules which will be included in the upcoming beta.

PR #39, opened 2018-04-08
Add ability for users to change custom file names

torkel104:
Use aliases for class config files instead of hardcoding file names. This way the user can choose their own file names instead of the now hardcoded 'class_c.cfg,' convention. For example this allows the user to avoid file clutter by putting class files in their own directory. As another example the files can be organised by class number instead of alphabetically by using file names like '1_scout.cfg.'


mastercoms (approved):

mastercoms:

PR #87, opened 2019-07-30
"Unknown command block_game_override_once" fix

Tiagoquix:
Fix for the "Unknown command block_game_override_once"


mastercoms:
The point is so that it doesn't run the first time but runs every time after that.

PR #90, opened 2019-08-01
Misc things

Tiagoquix:


config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -402,7 +402,7 @@ alias post_processing_ultra "mat_hdr_level 3;pp_pass_increment;pp_pass;mat_bloom
 //pyro_vignette_distortion 0 // Disable vignette distortion for pyrovision
 //pyro_vignette_distortion 1 // Enable pyrovision vignette distortion
 
-alias pyrovision_low "pyro_dof 0;pyro_vignette 0"
+alias pyrovision_low "pyro_dof 0;pyro_vignette 0;pyro_vignette_distortion 0"
```
mastercoms, line 405: Since pyro vignette is off, no need to set distortion.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -845,7 +845,7 @@ gl_paircache_ways_lg2 5 // Keep some loops small by using minimum ways size
 //tf_scoreboard_mouse_mode 1 // Always use mouse mode when scoreboard is open
 //tf_scoreboard_mouse_mode 0 // Always look around, even when the scoreboard is open
 //tf_scoreboard_alt_class_icons 1 // Alternate class icons on the scoreboard
-//tf_use_match_hud 1 // Show the top player bar on all servers
+tf_use_match_hud 1 // Show the top player bar on all servers
```
mastercoms, line 848: Match HUD is enabled by default. Why do you think the config should enforce this option?

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -860,7 +860,7 @@ tf_colorblindassist 1 // Add extra icon for jarate (colorblind mode)
 //cl_disablehtmlmotd 1 // Disable HTML MOTDs (ads, music player, rules, fancy server intros)
 sv_motd_unload_on_dismissal 1 // Server will tell the player to unload the MOTD
 //hud_classautokill 0 // No suicide on class change
-//tf_halloween_bonus_ducks_cooldown 10 // Delay in seconds before BONUS DUCKS sound again
+tf_halloween_bonus_ducks_cooldown 10 // Delay in seconds before BONUS DUCKS sound again
```
mastercoms, line 863: This is the default, why do you think the config should set this?

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -953,7 +953,7 @@ tf_time_loading_item_panels 0.0004 // Decrease time spent per frame loading item
 cl_spec_carrieditems 1 // Show spectated player items
 //cl_spec_carrieditems 0 // Do not show spectated player items
 tf_backpack_page_button_delay 0.25 // Decrease button delay for moving items
-//tf_item_selection_panel_sort_type 1 // Alphabetically sort loadout options
+tf_item_selection_panel_sort_type 1 // Alphabetically sort loadout options
```
mastercoms, line 956: People are used to their current sorting, and this shouldn't be the default.

mastercoms (changes requested):

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: I think the config should enforce this on because it's a useful option to leave it on. No need to turn it off.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Oh, I didn't know this was the default value.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Why not?

mastercoms: Because people are going to be used to how it's sorted in their inventories already, so it'll be weird if the config unexpectedly changed this. It is a hidden option that people can set themselves if they want to.

config/mastercomfig/cfg/comfig/comfig.cfg
mastercoms: Well, it's on by default, so it would only be off if the user turned it off. As such, I don't think we should be messing with a decision the user did.

Tiagoquix: Alright. Leave it commented.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Alright. Leave it commented.

PR #91, opened 2019-08-04
Make "Silly Launch Options" page consistent

Leo40Git:
The launch options are now inline code, whoopee. Also does -nosrgb even have a slight impact on performance or is it just something that Source has for some reason?


mastercoms (approved):

mastercoms:

It probably does, but it's unlikely that it'll make a huge difference on any sort of decent PC.

PR #98, opened 2019-08-10
sv_cheats 0

Tiagoquix:


mastercoms:
The config doesn't set sv_cheats 1 anymore, so the only time this is the case is if the user sets it.

PR #107, opened 2019-09-01
Modified medium preset.

Tiagoquix:


mastercoms:
This is intentionally skipping setting lod and picmip in order to use what's recommended for someone's system automatically.

PR #108, opened 2019-09-01
Modified medium-high preset.

Tiagoquix:


mastercoms:
Most modules that you're changing are intentionally skipped in order to use dxsupport instead.

PR #109, opened 2019-09-01
I guess it doesn't work without the double quotes.

Tiagoquix:


mastercoms:
This improves performance and memory usage. Spaces are not necessary for aliases though.

PR #117, opened 2019-09-02
Fix game_overrides_once not working

Tiagoquix:
Didn't test, but I think it works now.


mastercoms:
I'd rather not change the names, because that could break people's settings.

PR #118, opened 2019-09-02
Some mini-changes #2

Tiagoquix:
I removed the `alias download download_mapsonly` beucase I think people will prefer to set it via menu instead of using the `modules.cfg` file. This is useful for people that play on community servers, like Freak Fortress 2 servers.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -24,8 +24,8 @@ cl_smoothtime .1 // Use smoothing based on human tolerance
 hud_escort_interp .15 // Smooth out payload cart progress over slightly less time
 //cl_pred_optimize 1 // Avoid bug with mode 2 that causes client prediction to be used over server data
 
-alias pred_opt_full "cl_pred_optimize 2"
 alias pred_opt_base "cl_pred_optimize 1"
+alias pred_opt_full "cl_pred_optimize 2"
 
 // -----------------
 // '-- Snapshots --'
@@ -109,8 +109,6 @@ alias download_all "cl_downloadfilter all"
 alias download_nosounds "cl_downloadfilter nosounds"
 alias download_mapsonly "cl_downloadfilter mapsonly"
 alias download_none "cl_downloadfilter none"
-
-alias download download_mapsonly
 
 // -------------------
 // '-- Matchmaking --'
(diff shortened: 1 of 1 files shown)


config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -881,6 +879,7 @@ alias hud_player_model_on "cl_hud_playerclass_use_playermodel 1"
 
 alias dynamic_background_off
 alias dynamic_background_on "map_background background01"
+
```
mastercoms, line 883: No need for a line here.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -110,8 +110,6 @@ alias download_nosounds "cl_downloadfilter nosounds"
 alias download_mapsonly "cl_downloadfilter mapsonly"
 alias download_none "cl_downloadfilter none"
 
-alias download download_mapsonly
```
mastercoms: I set it by default because I haven't seen any complaints about it, but have seen complaints the other way (https://www.teamfortress.tv/52468/dm-server-jointime-fix).

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1097,14 +1096,14 @@ fov_desired 90 // See more of the battlefield
...
 //glow_outline_effect_enable 1 // Enable ally and objective x-rays
 //glow_outline_effect_enable 0 // Disable all x-rays
 //tf_enable_glows_after_respawn 1 // Enable spawn x-rays
 //tf_enable_glows_after_respawn 0 // Disable spawn x-rays
-//tf_spec_xray_disable 0 // Enable xrays while spectating
-//tf_spec_xray_disable 1 // Disable xrays while spectating
-tf_spectate_pyrovision 0 // Do not force use pyrovision in spectator mode
+//tf_spec_xray_disable 0 // Enable x-rays while spectating
+//tf_spec_xray_disable 1 // Disable x-rays while spectating
+tf_spectate_pyrovision 0 // Don't force pyrovision in spectator mode
```
mastercoms: I don't use quotation marks because it messes with some syntax parses which think it is the start of a string.

mastercoms (changes requested):

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: It's just to follow the rest of the aliases.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Alright.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Ok. But can you change the "xrays" to "x-rays" at least?

mastercoms:

PR #122, opened 2019-09-09
Update comfig.cfg

Tiagoquix:
I commented cvars that already are in their default value and mastercomfig doesn't have to forcing to the default value again.


mastercoms:

PR #125, opened 2019-09-11
Add cl_dynamiccrosshair again

Tiagoquix:


mastercoms:
It was removed because it doesn't work in TF2.

PR #127, opened 2019-09-16
Add hud_panels=off to very low preset

Tiagoquix:


mastercoms:
It's intentionally not set. We used to set this, and we got so many support requests about how to re-enable them that it's clear we shouldn't set it.

PR #135, opened 2019-10-02
Some cleanup

Tiagoquix:
We don't have to set tf_playergib, it's a server-side console variable. If anyone is playing in your own hosted server, that person won't see gibs if that person has it enabled.


mastercoms:
tf_playergib completely skips the gib setup code, while other commands just skip rendering or remove it more quickly. This is good for local server optimization.

PR #140, opened 2019-10-17
Update comfig.cfg

Tiagoquix:


mastercoms:
Did you test this?

Tiagoquix:
Yes. I always played with effects on Very Low setting and I did not get a crash when playing as Heavy.

PR #146, opened 2019-11-18
Miscellaneous changes

Tiagoquix:
For cl_ragdoll_fade_time: -1 isn't different from 0 in this specific cvar

For r_eyegloss: Idk if that makes a difference on FPS but here is it.

Please ignore dynamic_background_preload_room.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -460,8 +460,6 @@ alias texture_filter_aniso16x "mat_forceaniso 16"
 //r_eyes 0 // Disable eyes
 //r_eyemove 1 // Control if character eyes should move, does not really affect performance
 //r_eyemove 0 // Disable eye movement
-//r_eyegloss 1 // Enable eye gloss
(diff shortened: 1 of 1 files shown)


config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -461,6 +460,8 @@ alias texture_filter_aniso16x "mat_forceaniso 16"
 //r_eyes 0 // Disable eyes
 //r_eyemove 1 // Control if character eyes should move, does not really affect performance
 //r_eyemove 0 // Disable eye movement
+//r_eyegloss 1 // Enable eye gloss
```
mastercoms: r_eyegloss is not used

mastercoms (changes requested):

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: I didn't know about that. Ignore it then.

mastercoms (approved):

mastercoms:

PR #166, opened 2020-01-20
Fix incorrect values on Very Low preset

Tiagoquix:


mastercoms:
All the values here are intentional.

PR #168, opened 2020-01-20
Change effects module to low setting on medium preset

Tiagoquix:
I don't think medium setting on effects module would give the user an advantage.


mastercoms:
People have said in the past that the tracers can help them know where their bullets are going generally.

PR #181, opened 2020-02-04
Async loading breaks animations preloads after a few map changes

Tiagoquix:
Also, removing voice toggle commands, they are not necessary at all. The voice_enable command is bugged, if you toggle it during a match you will have to type retry into the console to re-enable the voice system after turning it on again.


mastercoms:
Should we block voice_enable again then?

Tiagoquix:
No.

mastercoms:
So why not have the shortcut to disable both, if we shouldn't block a bugged command, and we shouldn't have a shortcut to manage voice better.

Tiagoquix:
I guess many people will just type `voice_enable 0` into the console and that's it. There is no need for a shortcut, and there is no need to block the command.

mastercoms:
I feel like I'm going to hold off on these changes for now.

PR #188, opened 2020-02-07
Reworked mm_override.cfg

Tiagoquix:
Added more commands and fixed some values. Also the positioning is now alphabetically.


mastercoms:
Are these commands blocked by competitive?

Tiagoquix:
> Are these commands blocked by competitive?

Not sure, I'm adding some default values to various commands that mastercomfig changes. I'll update the mm_override.cfg now.

> Q: ARE THERE ANY MINIMUM REQUIREMENTS TO PLAY?
A: The only minimum requirement is DirectX 9. However, certain graphical settings will be enforced in order to ensure a level playing field. Players will only be able to modify graphics settings that are available in the Options and Advanced Options menus.

Source: http://www.teamfortress.com/meetyourmatch/faq/

mastercoms:
I can get the list of allowed cvars for you.

Tiagoquix:
> I can get the list of allowed cvars for you.

Alright.

mastercoms:
https://gist.github.com/mastercoms/c4b65d588a486cd58c765eccce284144

Tiagoquix:
Alright, what can I do? I am confused.

mastercoms:
Competitive mode uses a whitelist, rather than a blacklist, to stay on the safe side. The cvars in the files allowed-in-competitive.txt, hidden.txt, and internal-use.txt are allowed in competitive. They don't need to be reset. Then, in cvarlist.log, the cvars with GAMEDLL, ARCHIVE, REPLICATED, and CHEAT are not reset by competitive mode. So, everything else other than these cvars are restricted by competitive.

Finally, viewmodel_fov and mat_dxlevel have restricted values in competitive mode, but I don't think you need to set those.

Tiagoquix:
What about cvars with CLIENTDLL?

mastercoms:
They are reset if they do not have any of the other flags.

Tiagoquix:
I'm here editing the cvarlist.log file and I already found close caption cvars (`cc_`). It is strange that they are resetted.

mastercoms:
Not a surprise. TF2 does not have any captions by default, being a competitive multiplayer title. People can override their captions using things like [Clovervidia's captions](https://github.com/clovervidia/clovervidias-captions) to get a competitive advantage.

PR #190, opened 2020-02-07
Do not disable pause image

Tiagoquix:
It is useful, most used in community competitive servers.


mastercoms:
How does this affect the pausing? I think it only removes the PAUSED image in the middle of the HUD, right? And I think most competitive players would be aware of when the game is paused.

Tiagoquix:
> How does this affect the pausing? I think it only removes the PAUSED image in the middle of the HUD, right? And I think most competitive players would be aware of when the game is paused.

Yes, it only removes the image, but why would you need to remove that? It is a simple image that tells the user that the game is paused. Probably, some new player playing in a random server that gets paused would think that their PC has frozen or something like that.

sapphonie:
+1, removing is negligible on frames and has briefly confused me even though i've been playing this game for 3 years lol

mastercoms:
Because the GUI checks if the game is paused every frame. If it is, then the game shows the paused image. By skipping that, you can skip a few checks per frame, but it's a very small gain, so if people are confused, we can revert it.

mastercoms (approved):

PR #191, opened 2020-02-07
Null-cancelling Movement script addon

scrambl-d:


mastercoms:
How will we bind to the appropriate keys for movement?

scrambl-d:
> How will we bind to the appropriate keys for movement?

Default wasd are bound in the script, if you want to use custom wasd buttons you can bind them to the universally accepted +mfwd etc. from the null cancelling movement addon.

mastercoms:
It seems the null movement script is empty. I feel like if it requires that user specific customization just to set up, it should be left till later, when we have a way to be keybind agnostic.

scrambl-d:
> It seems the null movement script is empty. I feel like if it requires that user specific customization just to set up, it should be left till later, when we have a way to be keybind agnostic.

That's fair, I just assumed that the use case of having different wasd keys was so niche that you could assume they'd be willing to put in the legwork to make it work. If you're planning on having an app to manage keybinds it makes a lot of sense to wait until then. :)

mastercoms (approved):

mastercoms:

PR #211, opened 2020-02-24
Pyroland on all official maps

Tiagoquix:
Source: https://steamcommunity.com/sharedfiles/filedetails/?id=1879258346
I edited and added the maps Laughter and Precipice. I also removed spaces.


Tiagoquix:
Note: only official maps, none community maps are included.

mastercoms:
I included some popular competitive and creators.tf maps so that those modes could be better supported. Halloween maps were not added so that their style would not be changed.

PR #213, opened 2020-02-27
set snapshots_safe on load again

Tiagoquix:


Tiagoquix:
@mastercoms What is wrong with the check?

mastercoms:
Fixed

PR #216, opened 2020-02-27
Update comfig.cfg

Tiagoquix:
About fps_max: https://blog.counter-strike.net/index.php/2020/02/28730/

> – Increased competitive max fps limit to 400 to support higher refresh rate monitors.


mastercoms (approved):
Great work! Thanks!

PR #218, opened 2020-02-29
comfig.cfg update

Tiagoquix:
It is unnecessary to have the network_optimizations module at this point.
cl_interp_all and cl_SetupAllBones are defaulted to 0.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -63,7 +63,7 @@ alias hitboxes_safe "snapshot_buffer_safe;cl_pred_optimize 2;snapshot_buffer" //
 alias hitboxes_accurate "snapshot_buffer_test;cl_pred_optimize 1;snapshot_buffer" // Repredict for small changes in server-side values while still being reliable
 alias hitboxes_low_delay "snapshot_buffer_low;cl_pred_optimize 2;snapshot_buffer" // Minimize difference between interpolation and current server tick
 
-lias network_optimizations_off "cl_interp_all 1;cl_SetupAllBones 1"
+alias network_optimizations_off "cl_interp_all 1;cl_SetupAllBones 1"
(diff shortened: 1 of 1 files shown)


mastercoms:
Yes, but the user should be able to select the "off" option in order to get better hit registration possibly, at the cost of performance.

Tiagoquix:
> Yes, but the user should be able to select the "off" option in order to get better hit registration possibly, at the cost of performance.

So what do you think of it being in the "off" option by default and only "on" by default on the very low preset?

mastercoms:
It's significantly lower performance for a very slight increase in accuracy. I don't think it should be the default.

Tiagoquix:
> It's significantly lower performance for a very slight increase in accuracy. I don't think it should be the default.

If it is significantly lower performance, then I do not agree to have the option to choose "off" or "on". They always was defaulted to 0 on both values, and in most Source games this is also the default. I think we shouldn't mess with this.

mastercoms:
It used to be part of the experimental addon, but now is being moved into the base config.

Tiagoquix:
> It used to be part of the experimental addon, but now is being moved into the base config.

I understand, but I suggest to only document this on the main comfig.cfg file, with the appropriate commands.
Enforcing they to its defaults it is useless at this point.
As you wrote in the comfig.cfg: "Not intended to be documented at the moment.".

mastercoms:
Sometimes, modules set things to their default values, but only because another module level sets them to a different value. This is one of those cases. It's not documented at the moment because it needs further testing on my end, until the release.

Tiagoquix:
> Sometimes, modules set things to their default values, but only because another module level sets them to a different value. This is one of those cases. It's not documented at the moment because it needs further testing on my end, until the release.

Alright. Document it when you find the time is appropriated.
And about this pull request, will you merge it?

mastercoms:
I can take some of the changes. Will merge it soon.

mastercoms (dismissed):

mastercoms (approved):

PR #222, opened 2020-03-04
Spy interp/snapshot buffer

Tiagoquix:
That was wrong, I did only the basic and removed the values for you setting these later.


mastercoms:
Sorry, I'm not sure what this is for.

Tiagoquix:
> Sorry, I'm not sure what this is for.

Following the comfig style, `interp_standard/congestion_*` should not set `cl_interp_ratio`, and `snapshot_buffer_spy` should set the `cl_interp_ratio` value.

Tiagoquix:
I did only made the base for you to finish later.

mastercoms:
It should in this case, as congestion does not need a lower interp to be safe for server framerate drops.

PR #223, opened 2020-03-04
Spy network settings now follows comfig style

Tiagoquix:
https://github.com/mastercoms/mastercomfig/pull/222#issuecomment-594794438

Following the comfig style, I think it should be like this.


Tiagoquix:
This is not necessary at all, it is only a aesthetic issue.

mastercoms:
I don't understand the benefit of this.

Tiagoquix:
> I don't understand the benefit of this.

It has no benefit, it's just an aesthetic issue like I said. It's just to make it more "organized".

mastercoms:
I feel like it's more confusing, because of the additional alias layer.

PR #229, opened 2020-03-12
Presets updates

Tiagoquix:
I tweaked it a bit so that each preset really matches its description.


Tiagoquix:
I also did remove some unnecessary characters from some files. The same what @micwoj92 did, but with more files.

micwoj92 (commented):
In surface properties you can also remove quotes

micwoj92:
Also I would left notes, or  at least move them to readme, so someone who needs it, can edit that

Tiagoquix:
> Also I would left notes, or at least move them to readme, so someone who needs it, can edit that

If someone wants to edit, I think the person would read the original file.

PR #232, opened 2020-03-12
Removed unnecessary spaces on mtp.cfg

Tiagoquix:


micwoj92:
I think these are tabs and are there for visibility, imo better to make all equal in length

Tiagoquix:
> I think these are tabs and are there for visibility, imo better to make all equal in length

Yep, they are tabs. They are for visibility but the file isn't edited so much so they are not necessary to stay on the file. Plus, removing the tabs reduces the file size.

micwoj92:
Maybe then you should also remove quotation marks, these have no effect if there are no spaces in filenames

Tiagoquix:
> Maybe then you should also remove quotation marks, these have no effect if there are no spaces in filenames

They are present in the original file.

micwoj92:
I just checked to be 100% sure, with 0's and 1's, with and without quotes. All do work, that means quotation marks are not necessary.

Tiagoquix:
> I just checked to be 100% sure, with 0's and 1's, with and without quotes. All do work, that means quotation marks are not necessary.

I know about this. I am just keeping the quotation marks because they are present in the original file. You can make a pull request to remove it, if you wish.

micwoj92:
Then you should be consistent and also keep tabs, because they are present in the original file.

Tiagoquix:
> Then you should be consistent and also keep tabs, because they are present in the original file.

Nope, I prefer to remove the quotation marks too.

mastercoms:
Taking #232 instead

PR #235, opened 2020-03-13
Removed unnecessary quotation marks from smaller files

micwoj92:


mastercoms:
I feel like I will do a generative system for this instead, since the diff on these TF2 files is too big imo.

PR #242, opened 2020-04-12
fix typo

ghost:


mastercoms:
Hi, thanks for the contribution, but this is actually intended.

> **niche, adj.** denoting or relating to products, services, or interests that appeal to a small, specialized section of the population.

PR #251, opened 2020-04-18
Wipe up mm_override.cfg for testing

Tiagoquix:
I have an idea: wipe up mm_override.cfg and see if there will be any crash feedback. If not, there will no longer be any need for commands here, since TF2 itself changes the values ​​and people usually restart the game to remove the commands' values locks.
I know this exist for the purpose of avoiding crashings while loading, but a test would be good.


mastercoms:
People can just not run it if they want.

PR #252, opened 2020-04-18
Mention autoexec.cfg that is created automatically

Tiagoquix:


mastercoms:
It won't cause conflicts, as VPKs will always override the loose files in tf. But it may be good to say that it's not used, so users can clean it up.

Tiagoquix:
> It won't cause conflicts, as VPKs will always override the loose files in tf. But it may be good to say that it's not used, so users can clean it up.

Alright, I will add another commit.

mastercoms (approved):

PR #256, opened 2020-04-28
Do not block game_overrides_once in benchmarks

Tiagoquix:
block_game_overrides_once will not exist if the user does not enter a match before running it. Also, there is nothing on the current game_overrides_once alias, so blocking it is useless.


mastercoms:
Fixed in https://github.com/mastercoms/mastercomfig/commit/21ed331a3f6d82e86ab19007180448716f570cfb.

PR #260, opened 2020-05-01
Update comfig.cfg

Tiagoquix:


mastercoms:
After further examination, mod_forcedata should be toggled for mod_support.

Tiagoquix:
> After further examination, mod_forcedata should be toggled for mod_support.

Why?

mastercoms:
Because it forces all async tasks to finish, and waits. That should not be done, if we want the full benefit of async loading.

PR #262, opened 2020-05-01
launch_options.md

Tiagoquix:


mastercoms:
macOS doesn't have a registry, so it has to retain the dxlevel value.

Tiagoquix:
well, something is wrong

image

Tiagoquix:
@mastercoms

PR #263, opened 2020-05-02
launch_options.md

Tiagoquix:
My idea here is to agroup "Windows" and "macOS/Linux" together, and don't put "-dxlevel 100" into the info card. I don't know if the formatting is correct, but I think it is.


mastercoms:
Admonition doesn't wrap things like that, unfortunately.

PR #264, opened 2020-05-03
Comfig updates

Tiagoquix:


config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -28,38 +28,37 @@ net_queued_packet_thread 0 // We do not split on the client
 net_splitrate 0 // Skip trying to process queued split packets every tick
 net_udp_rcvbuf 8192 // Receive buffer should only be large enough for processing one frame
 net_chokeloop 1 // Early check for multiplayer
-sv_lan 1 // Protect local servers
```
mastercoms: Why remove `sv_lan 1`? People are reporting that hackers are joining their local games.

mastercoms (changes requested):

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: A local game is only joinable if you did "Port Forwarding". Otherwise, it is not.

mastercoms: What about UPnP? It still happened, and a user complained about it. So it's good for safety, now that hackers are joining local servers.

mastercoms (approved):

config/mastercomfig/cfg/comfig/comfig.cfg
CrescentSpirit: I can confirm this happened to me personally; two random people joined my local/listen itemtest server out of nowhere. Didn't give them time for any griefing by disconnecting and immediately proceeded to enable the sv_lan cvar. Haven't had any trouble since. Also, I have been using mastercomfig's default network configuration and have never messed with any "Port Forwarding" settings so I'm not sure if that option is actually relevant. In any case, given the circumstances and plentiful evidence, having sv_lan enabled is essential in my opinion, and defaulting to it was a very good (preemptive) move by mastercoms; will save a lot of people from needless scares and confusion 👍

PR #265, opened 2020-05-04
Comfig updates

Tiagoquix:


Tiagoquix:
"finalize.cfg // add host_writeconfig because mat_antialias is **alto** present in the user's config.cfg"
I mean "**also**", sorry

docs/customization/modules.md
```diff
@@ -52,7 +52,8 @@ Controls how fast you send and receive from the server.
 Default setting: **`packet_rate=standard`** (all presets, except Very Low).
 
 * **`packet_rate=congestion`**: A halved (33) packet rate for congested networks, usually bad WiFi, or very slow CPUs.
-* **`packet_rate=standard`**: The standard 66 packet rate which is the maximum for all servers.
+* **`packet_rate=standard`**: The standard 66 packet rate which is the maximum for all normal servers.
+* **`packet_rate=max`**: The maximum 100 packet rate which is supported by some modded servers.
```
mastercoms: I don't want packet rate max to be documented, because people will use it for no reason.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -358,8 +358,8 @@ alias particles_ultra "tf_impactwatertimeenable 0;fx_drawimpactdebris 1;fx_drawi
 //mat_autoexposure_max 2 // Max autoexposure value
 //mat_autoexposure_min .05 // Min autoexposure value
 
-alias post_processing_off "mat_hdr_level 0;mat_colorcorrection 0;mat_colcorrection_disableentities 1;mat_autoexposure_max 0;mat_autoexposure_min 0;r_bloomtintr 0.333;r_bloomtintg 0.333;r_bloomtintb 0.333;r_bloomtintexponent 1"
```
mastercoms: It shouldn't be set to 1, because it's controlled by lighting module. The higher post processing modules just enable it as well.

mastercoms (changes requested):

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: And why the others levels have it set to 0?

mastercoms: Because they need it, it disables the post processing effects on 1. But it's turned on/off in the lighting module. So low post processing does not depend on its value, but higher ones do.

Perkzitos: It should be removed only from post_processing_off or from post_processing_low? Or from the two?

mastercoms (approved):

PR #279, opened 2020-05-21
Improve Troubleshooting page

Leo40Git:
- Rewrite "crash on joining Competitive match" section
- Make some sentences less awkward
- Add some newlines here and there
- Bunch up the two "ignore these errors" sections
- Add TL;DR to "change interp value" section

### DIFF
--- a/docs/next_steps/troubleshoot.md
+++ b/docs/next_steps/troubleshoot.md
@@ -40,12 +40,12 @@ alias game_overrides_once_c hud_reloadscheme
 
 The command will only run on the first time you spawn in a match.
 
-## TF2 crashes when entering a Competitive match
+## TF2 crashes when entering a Competitive Mode match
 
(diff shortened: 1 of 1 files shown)


docs/next_steps/troubleshoot.md
```diff
@@ -30,40 +30,44 @@ Instead of `70` for `viewmodel_fov`, you can use any value you prefer.
...
 
 ```c
 alias game_overrides_once_c hud_reloadscheme
 ```
 
-The command will only run on the first time when you spawn in a match.
+The command will only run on the first time you spawn in a match.
 
-## TF2 crashes when entering a Competitive Matchmaking match
+## TF2 crashes when entering a Competitive match
```
micwoj92: I think this could be "Competitive Mode match" to prevent confusion between competitive and in-game competitive.

Leo40Git: I don't think anyone would make that mistake, but done.

mastercoms (approved):

PR #281, opened 2020-05-24
Mention -arbmode launch option

felikcat:


mastercoms:
Could you detail your findings with -arbmode to back up the description? (here, not in the contribution, so I can know whether to approve it or not)

felikcat:
Routes kGLM* functions through ARB instead of GLSL.
ARB is a faster method, but supports less OpenGL features; I haven't seen graphical rendering mishaps from using it for TF2.
It seems NVIDIA specific with the improvements, `amdgpu` driver is margin of error, but there's no code to show that it's ignored for non-NVIDIA.

`-glmdualshaders` launch option is supposed to switch between kGLMGLSL and kGLMARB in-game if you hold down the Shift key, but has its keybinding disabled.

docs/customization/launch_options.md
```diff
@@ -62,6 +62,8 @@ Read below about optional launch options and choosing your own DXLevel.
 
 === "Linux"
     * **-console** : displays console on startup
+    * **-arbmode** : set ToGL's drawing language to ARB instead of the default GLSL, less frame variance and higher FPS on NVIDIA; forces fast [clipping](https://developer.valvesoftware.com/wiki/Clip_texture), which may cause graphical issues.
```
mastercoms: Can you put the two launch options together, if they're meant to be used together?

mastercoms: That fast clipping link is not accurate, please remove it.

mastercoms: Software clipping may cause performance issues as well.

mastercoms: Change fast clipping to software clipping, so that the name is not misleading to average users.

mastercoms: I think it should read `set ToGL to use ARB instead of the default GLSL`.

mastercoms (changes requested):

mastercoms (approved):

mastercoms:
Just a heads up, this actually doesn't do anything, because `-glslcontrolflow` forces GLSL on. So if `-arbmode` didn't work without it, that means ARB mode doesn't work, and you were just going back to GLSL.

PR #291, opened 2020-05-25
Addons updates

Tiagoquix:
Following the layer:

config
presets
addons
user

@mastercoms


micwoj92:
these long lines look horrible, but noone reads it so whatever

mastercoms:
Thanks for the idea. I will restructuring addons to better support this, and include it in the next version. Will write you as a co-author.

PR #292, opened 2020-05-29
comfig.cfg updates

Tiagoquix:


config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1282,7 +1283,7 @@ snd_mixahead .066 // Balance between sound delay and less demand on the CPU. Syn
 // ----------------------
 // Adjust spatialization (sound travel and volume falloff) and DSP effects (sound filters)
 
-//dsp_enhance_stereo 0 // Disable extra spatial DSP effects and delays
```
mastercoms: I don't want to force this off, since it is automatically applied with high quality sound when using headphones. Hard to detect for mastercomfig, but easy for the game.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -742,10 +743,10 @@ r_occlusion 1 // Use CPU to have the GPU skip rendering models/props you cannot
 r_occludeemaxarea 1 // Skip occlusion of objects that are too large to test for
 r_occluderminarea 1 // Skip occluders that are too small to block anything worth our time
 r_occludermincount 1 // Just in case we filter one out
-//mat_tonemapping_occlusion_use_stencil 0 // Force disable stencil buffer for tonemapping
```
mastercoms, line 745: I don't want to force this off, since it is automatically determined based on the graphics card.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -634,13 +635,13 @@ alias props_ultra"r_decalstaticprops 1;cl_phys_props_enable 1;r_drawdetailprops
 //cl_ragdoll_physics_enable 0 // Disable ragdoll physics, where most of the performance from ragdolls comes from
 //cl_ragdoll_collide 1 // Enable ragdoll collisions
 //cl_ragdoll_collide 0 // Disable ragdoll collisions
-//ragdoll_sleepaftertime 1.5 // Wait a reasonable time before sleeping ragdoll physics
```
mastercoms, line 637: I don't think we need to set this with f.

mastercoms (changes requested):

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: https://www.reddit.com/r/GlobalOffensive/comments/3zqtvm/improved_csgo_sound_why_the_popular_settings_suck/

> **dsp_enhance_stereo** is supposed to 'enhance the stereo effect' for a slight performance hit when set to "1". In general, because digital effects that emulate stereo mixing are a bad idea as they to alter the apparent left-right positioning of sounds, and because it took extra CPU workload, I would set it to "0". But in some limited testing I did before writing this, I couldn't tell any appreciable difference between the two settings, so I suspect Valve has made them the same so that one does not offer a competitive advantage.

mastercoms: I notice a difference, and also that's CSGO, which has a different audio engine at this point.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: This is not the unique cvar with a `f` on it.
`tf_halloween_kart_*` are also cvars with a `f` on it. By this logic, I conclude that the `f` is there for a reason.

mastercoms: I don't think so, I have looked into it and there's no parsing difference since they are always parsed as floats.

Tiagoquix: > I don't think so, I have looked into it and there's no parsing difference since they are always parsed as floats.

So why the letter `f` is present?

Tiagoquix: I will add the letter `f`.

mastercoms: Sorry, I thought I responded to this. Probably a stylistic choice or misunderstanding by the developer.

Tiagoquix: I am doing another PR. I closed this but I will keep my changes here into the another PR.

PR #296, opened 2020-06-01
Addons hotfix and config changes

Tiagoquix:
**Changes:**

- Bug Report template issue on GitHub was updated (minor change).
- Addons have been updated to make they work correctly now.
- All addons were moved to `config/addons/mastercomfig-addonname-addon`, including all the `.cfg` files.
- All presets were moved to `config/mastercomfig/cfg/presets`.
- Minor comfig.cfg updates, forced some defaults, updated commentaries and fixed lod levels (medium preset now uses lod=low).
- Added the letter `f` to `ragdoll_sleepaftertime`, since the default value has it.

**Needed changes:**

The files:

`dev/docs/addons/package.sh`
`dev/docs/comfig/package.sh`
`dev/docs/presets/package.sh`
and probably `dev/docs/release.sh` need to be changed.

Since I moved the files, somethings will break. 
Also, the first addon run should be renamed to `addons_firstrun` and the second addon run should be renamed to `addons_secondrun` (`dev/docs/presets/package.sh`). I do not know how to mess with dev files, and I will not mess with them, since I can make something wrong in the process. I will let you do it @mastercoms, if possible.

Also, where is the `exec user/modules.cfg` in `dev/docs/presets/package.sh`?


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -17,7 +17,7 @@ Tell us the steps on how to recreate the bug.
 If applicable, add screenshots to help explain your problem.
 
 **Desktop (please complete the following information):**
```
micwoj92, line 19: Not really related to your changes, but maybe it should be renamed to just computer? Or like "Please fill in the following information"

Tiagoquix: Nah, "Desktop" is good.

mastercoms:
Changes that I wanted to include have been included in 8.102.1. Thank you for your contribution to mastercomfig.

PR #309, opened 2020-06-10
Add logo as module

micwoj92:
This has worked as standalone cfg file. I hope it doesnt break anything and if it would it would only print logo to console.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -1344,6 +1344,38 @@ alias voice_chat_on"voice_enable 1;voice_modenable 1"
 
 alias voice_chat voice_chat_on
 
+// ------------------
+// '----- Logo ------'
+// ------------------
+
+alias logostart"echo"                |||||"; l1;"
+alias l1"echo"           ||||||||||"; l2;"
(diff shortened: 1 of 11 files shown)


mastercoms (changes requested):
You don't need the logo=on for each config if you have https://github.com/mastercomfig/mastercomfig/pull/309/files#diff-7d84ef486bdbd2151929cf5f40299805R1377

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1619,35 +1652,8 @@ block_game_overrides_once
 
 alias version_comfig"echo mastercomfig version: 8.102.3 | June 8 2020"
 
+logo
```
mastercoms: Comfig is executed before modules. Perhaps we can move the echo to finalize.cfg?

micwoj92: Sorry I had no idea, I just added it above the current echos so it would echo in right place. I do not know which is better.

micwoj92: Moved

micwoj92: Do you mean to move the old echo as well?

PR #310, opened 2020-06-10
remove logo

micwoj92:
So this works as standalone echo, but doesnt as vpk because the packaging scripts script deletes comments and as @Perkzitos pointed out, the line 1647 didnt work after packaging.
I will play around with it and try to make it as working module, sorry for mistake.


Tiagoquix:
I didn't say it didn't work. I say that may be it won't work.

TIP: test the echo.

micwoj92:
I tested. It worked as standalone echo, but the packaging script strips comments so it removed it.

mastercoms (approved):

PR #312, opened 2020-06-12
Add logo as module attempt 2

micwoj92:
This moves the echo to separate file so it  doesn't get stripped by script. I tested vpk generated after these changes with windows and gnu/linux using very-low and low preset as well with logo=off and logo=on in modules.cfg with positive outcome.


micwoj92:
I don't get this bot

mastercoms (approved):

PR #314, opened 2020-06-17
Remove os/index.md

micwoj92:
In my opinion this page is unneeded and "OS Optimizations " repetition does not look good stylistically.
image


mastercoms:
My concern is having a centralized location to link to for people with different operating systems. And if we add new operating systems in the future, we have to update with more and more links in all the respective places rather than having a central location for them.

micwoj92:
Is it possible to "link" it to folder in repo rather than file? So for example linking it to  docs/os/ would give a page with Windows and Linux link. And it would highlight the above "OS Optimizations" text rather than the subfolder one
image

mastercoms:
No I don't think it's possible.

PR #316, opened 2020-06-17
Add remark-lint-checkbox-content-indent

micwoj92:
I think this should fix the error found in this log https://sider.review/gh/repos/69422496/pulls/315/logs/e4395b09-a8e2-4f67-b7ca-f97db5df89b8


mastercoms:
I don't think this change fixes the error. I will look into it, thanks.

PR #331, opened 2020-06-25
Update modules

micwoj92:
My reasoning is that now there is mod support module, map background has much more to do with it than with hud modules.


mastercoms (changes requested):
As long as `dynamic_background=dustbowl` exists, I don't think this is a change we want to do. We could improve the linking/formatting for the dynamic_background module though.

micwoj92:
The modules are next to each other in docs so this isn't big change, but I find the description of mod support to be confusing. I think it would be better to add something like "If you find your mods not working you can enable mod support module" to map support module and delete "You may want to also enable preloading using the Map Background module." from mod support description

mastercoms:
I'll improve the description then instead of moving it.

PR #333, opened 2020-06-26
Add info about user folder and where to find modules.cfg

micwoj92:


docs/customization/modules.md
```diff
@@ -6,14 +6,18 @@ description: Easily customize mastercomfig to get the exact stuff you want witho
...
+Modules are used in presets to set quality levels and can also be used in the `tf/cfg/user/modules.cfg` file to customize your config easily without searching for individual console variable values. You can find out what modules does your preset use [here](https://github.com/mastercomfig/mastercomfig/tree/release/config/cfg/presets). To set custom module value just put it in `user/modules.cfg`. For example:
 
 ```
 textures=high
 ```
 
 This will set the texture quality to `high`.
 
+!!! note
+    If you do not have `user` folder you should create it. Template `modules.cfg` file can be downloaded [here](https://github.com/mastercomfig/mastercomfig/releases/latest/download/modules.cfg)
```
mastercoms: I don't think the template should be mentioned here.

docs/customization/modules.md
```diff
@@ -6,14 +6,18 @@ description: Easily customize mastercomfig to get the exact stuff you want witho
 Modules are groups of related commands with specific values for a certain use case or quality level.
 
 ## How to use
-Modules are [used in presets](https://github.com/mastercomfig/mastercomfig/tree/release/config/cfg/presets) to set quality levels and can also be used in the `tf/cfg/user/modules.cfg` file to customize your config easily without searching for individual console variable values. Just put your desired module level in `user/modules.cfg`. For example:
+
+Modules are used in presets to set quality levels and can also be used in the `tf/cfg/user/modules.cfg` file to customize your config easily without searching for individual console variable values. You can find out what modules does your preset use [here](https://github.com/mastercomfig/mastercomfig/tree/release/config/cfg/presets). To set custom module value just put it in `user/modules.cfg`. For example:
```
mastercoms: I think it should be 
```
can also be used in your very own `tf/cfg/user/modules.cfg` file
```

mastercoms: I'm not sure if I like `You can find out what modules does your preset use [here](https://github.com/mastercomfig/mastercomfig/tree/release/config/cfg/presets)` over the previous way it was linked.

mastercoms: I think this works better:

```
To set a custom module level just put it in `user/modules.cfg`. For example:
```

mastercoms (changes requested):

docs/customization/modules.md
micwoj92: Should it be in customization instead? it's just so people wont make mistake of having modules.cfg.txt when creating manually

docs/customization/modules.md
micwoj92: done

micwoj92: I wanted to specifically add info that you can find default values for presets there. I dont know how to put it in another words.

micwoj92: done

docs/customization/modules.md
mastercoms: If that's the case, then an empty modules file should be linked.

docs/customization/modules.md
mastercoms: Ok, then fix the grammatical error, plus [use descriptive text](https://web.dev/link-text/) for the link: `You can find out what module levels your preset uses by [viewing their respective cfg file](https://github.com/mastercomfig/mastercomfig/tree/release/config/cfg/presets).`

docs/customization/modules.md
micwoj92: This points to template so users already can see module values and read the inside of file for instruction instead of going back and forth to the docs page. If link would be to emply modules.cfg then it also should be in my opinion generated by packaging scripts. I do not believe this change is needed and the template is much better option.

mastercoms: The template encourages people to go through each option which is non-ideal and also involves editing a text file, rather than creating your own layer on top of mastercomfig as intended. This creates harder upgrade paths when modules change, as it involves redoing your modules.

docs/customization/modules.md
micwoj92: done

docs/customization/modules.md
micwoj92: Ah, I understand that point.  I see now that creating modules.cfg as empty file (same as autoexec.cfg) and renaming current modules.cfg to something like templatemodules.cfg as best option. Not sure about what you think about that.

mastercoms: You can remove the sentence for now, and then we can add it back once we have a generator for it.

docs/customization/modules.md
```diff
@@ -16,7 +16,7 @@ textures=high
 will set the texture quality to `high`.
 
 !!! note
-    If you do not have `user` folder you should create it. Template `modules.cfg` file can be downloaded [here](https://github.com/mastercomfig/mastercomfig/releases/latest/download/modules.cfg)
+    If you do not have `user` folder you should create it. Same principle applies to `modules.cfg` file.
```
micwoj92, line 19: Edited the sentence

mastercoms (approved):

PR #345, opened 2020-07-02
 Prevent fog cvar spew in console on startup

micwoj92:
This was present in old versions of mastercomfig but it was removed later. I am open for discussion as I do not know reason why it was removed. I think enabling it to prevent error messages and then disabling later in cfg is a reasonable fix for the problem that is cheat protected commands executing and giving errors.


mastercoms:
Are you sure that the fog cvars don't run after finalize?

micwoj92:
I will check it later today, thanks.

mastercoms:
Fog cvars run after, so it won't work.

micwoj92:
Oh I thought it's running when joining map using preload module. Thanks for correction and info.

PR #360, opened 2020-07-08
Make "cl_downloadfilter all" the default setting.

xNanochip:
This pull request makes the client cvar "cl_download filter all" the default module setting. The reason for this change is to eliminate the confusion on why clients get a giant "ERROR" for when they do not have required assets on a community server. This ensures that their game is running as intended and avoids obstructing game-play by not having the annoying "ERROR" sign.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -174,7 +174,7 @@ alias download_nosounds"cl_downloadfilter nosounds"
 alias download_mapsonly"cl_downloadfilter mapsonly"
 alias download_none"cl_downloadfilter none"
 
-alias download download_all
+alias download
 
 // -------------------
 // '-- Matchmaking --'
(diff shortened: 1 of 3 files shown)


Leo40Git:
The default should probably be `nosounds` if you're going for avoiding obstructive elements.

micwoj92:
Keeping in line with the new approach to mastercomfig defaults, the default for the downloads module should not be set.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -174,7 +174,7 @@ alias download_nosounds"cl_downloadfilter nosounds"
 alias download_mapsonly"cl_downloadfilter mapsonly"
 alias download_none"cl_downloadfilter none"
 
-alias download download_mapsonly
+alias download download_all
```
mastercoms:
```suggestion
alias download
```

mastercoms (changes requested):

mastercoms (changes requested):
You should change the documented default in `modules.md`.

xNanochip:
I assume you mean modules.cfg?

mastercoms:
In addition to the diff you have in modules.cfg, every module has their default documented in modules.md. With the default change, modules.md needs to be updated.

xNanochip:
Okay, good to go?

mastercoms (approved):

PR #378, opened 2020-08-12
setinfo

KaineWooten:
Set for the alias in comfig for console automcomplete


mastercoms:
setinfo adds overhead to the networking system, so I will not be taking this change.

PR #381, opened 2020-08-18
Add AZERTY version of null cancelling movement addon

treacherousfiend:
User on the discord needed a version for their AZERTY keyboard, so I decided to create a PR for it.

Contains files for the actual addon, as well as updating the docs to acknowledge its existence
Should build correctly, but I can't test since i've never been able to get the build scripts to detect vpk.exe

Perhaps when the installer comes back this can be removed and the installer can either detect the movement keys from the game or ask the user to input their movement keys.


mastercoms:
Would probably be better to reuse the other version, and just have two versions of the binding commands, through somewhat of a module.

treacherousfiend:
isn't that basically what I did?
I think the only other way would be for the user to build their own vpk, which i'm sure most people would like to not do

treacherousfiend:
alright i'll close the PR and rewrite using modules instead

mastercoms:
Not modules necessarily, but an alias based system like it.

treacherousfiend:
yeah i've been trying to figure out the least hacky way of doing it

PR #395, opened 2020-10-03
Outlines module changes

Tiagoquix:
The low level was activating more outlines/nametags than the medium level. To correct that, low level now only enables objective outlines and medium level only disables spectator outlines. I also changed the outlines module to be at the medium level by default on the medium and medium-high presets.


mastercoms:
This is an invalid change which does not meet the requirements of each module level.

Please refer to the modules documentation which clearly details the purpose of these two levels.

PR #399, opened 2020-10-05
Minor changes to the config

Tiagoquix:


docs/customization/modules.md
```diff
@@ -17,7 +17,7 @@ textures=high
 will set the texture quality to `high`.
 
 !!! note
-    If you do not have `user` folder you should create it. Same principle applies to `modules.cfg` file.
+    If you do not have `user` folder you should create it. Same principle applies to the `autoexec.cfg` file.
```
mastercoms: This is on the modules page. It's saying if you don't have modules.cfg, create it, just like if you don't have a user folder, create it.

docs/customization/modules.md
```diff
@@ -198,13 +198,13 @@ Default setting: based on which preset you are currently using.
 * **`particles=high`**: Enables weather particles, barely reduces particle density, enables bullet impact flecks.
 * **`particles=ultra`**: Enables weather particles, uses full particle density, enables the new bullet impact effects system, enables bullet impact flecks.
 
-### General Post Processing
+### General Post-Processing
```
mastercoms, line 200: I do not agree that there should be a dash here.

docs/customization/modules.md
```diff
@@ -345,7 +345,7 @@ Default setting: **`sillygibs=auto`**.
 
 Controls the rendering of various small objects.
 
-Keep in mind that you must add your preferred `cl_detaildist` and `cl_detailfade` settings to your class configs if they are different from your preset, or else they will be overridden.
+Keep in mind that you must add your preferred `cl_detaildist` and `cl_detailfade` settings to your class configs if they are different from your preset, or else they will be overridden. You probably won't change these console variables' values, and if you don't want to, you don't need to, since mastercomfig sets the appropriated values automatically.
```
mastercoms: Just remove this section, it's completely useless.

docs/customization/modules.md
```diff
@@ -103,7 +103,7 @@ Filters what custom content is allowed to be downloaded from the server.
 
 Default setting: **`download=auto`**.
 
-* **`download=auto`**: Whatever the user set in game settings.
+* **`download=auto`**: Whatever the user set in the game settings. By default, the game downloads every custom thing from community servers.
```
mastercoms: The game default is to download all custom files from the server.

docs/customization/modules.md
```diff
@@ -53,11 +53,11 @@ Default setting: **`packet_rate=standard`** (all presets, except Very Low).
 ### Snapshot Buffer
 
 Controls how reliable to consider the snapshots you receive from the server to apply client-side buffering/smoothing between snapshot receives.
-Also known as linear interpolation (lerp or interp). Note: mastercomfig sets this automatically per class, but setting it in your `modules.cfg` will override this behavior.
+Also known as linear interpolation (lerp or interp). Note: by default, mastercomfig sets this automatically per class, but setting it in your `modules.cfg` will override this behavior.
```
mastercoms: Remove this note, since it's covered in auto.

docs/customization/modules.md
micwoj92: above and below text is about modules, why change it then?

docs/customization/modules.md
```diff
@@ -53,11 +53,11 @@ Default setting: **`packet_rate=standard`** (all presets, except Very Low).
...
 
 Default setting: **`snapshot_buffer=auto`** (all presets).
 
-* **`snapshot_buffer=auto`**: Automatically
+* **`snapshot_buffer=auto`**: Automatically sets the best interpolation value based on which class you are currently playing.
```
mastercoms: Automatically set per class.

mastercoms (changes requested):

docs/customization/modules.md
Tiagoquix: Yes, that's what I explained.

docs/customization/modules.md
Tiagoquix: I will remove it.

docs/customization/modules.md
Tiagoquix: You don't have to. The correct grammar is with a dash.

Tiagoquix:
I moved the `dynamic_background` module run to `finalize.cfg`. It's best to run it after everything has been set.

docs/customization/modules.md
Tiagoquix: @mastercoms https://english.stackexchange.com/questions/452969/pre-processing-post-processing-hyphen-presence-in-research-publications-writ

docs/customization/modules.md
```diff
@@ -57,7 +57,7 @@ Also known as linear interpolation (lerp or interp). Note: mastercomfig sets thi
 
 Default setting: **`snapshot_buffer=auto`** (all presets).
 
-* **`snapshot_buffer=auto`**: Automatically
+* **`snapshot_buffer=auto`**: Automatically sets the best interpolation value per class.
```
mastercoms, line 59: Remove the note above: "Note: mastercomfig sets this automatically per class, but setting it in your `modules.cfg` will override this behavior."

mastercoms (changes requested):

Tiagoquix:
Done and done
@mastercoms @micwoj92

docs/faq.md
```diff
@@ -102,11 +102,11 @@ Class configs are also a reliable way to periodically run some commands during a
 * Every path in the game can only have one file associated with it, due to how Source handles mods
 * Drag and drop installation, especially for those who do not want to mess with configs
 
-## Why is there no mastercomfig app?
+## Why is there no mastercomfig application?
```
mastercoms: Use app instead of application. The branding for mastercomfig 7 was always "app".

mastercoms (changes requested):

Tiagoquix:
Done @mastercoms

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -17,7 +17,7 @@ Tell us the steps on how to recreate the bug.
 If applicable, add screenshots to help explain your problem.
 
 **Desktop (please complete the following information):**
-* Config version: don't say latest, say the actual version, found by running `version_comfig` in console
+* Config version (don't say latest, say the actual version, found by running `version_comfig` in console):
```
mastercoms:
```suggestion
* Config version: <!-- don't say latest, say the actual version, found by running `version_comfig` in console -->
```

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -104,12 +104,13 @@ alias packet_rate packet_rate_standard
 //cl_interp_ratio 2;cl_interp .030303 // 2 snapshots interpolation buffer to avoid issues where one snapshot is dropped once in a while. Safest option for most users.
 //cl_interp_ratio 3;cl_interp .045454 // Protect more against packet loss, increase further for more buffer protection. Recommended only for users with an unstable connection.
 
+alias snapshot_buffer_auto"alias snapshot_buffer"
```
mastercoms: I would rather use the shortcut for auto than a full module level.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -170,12 +171,13 @@ net_maxfilesize 64 // Max out file upload size for extra content
 //cl_downloadfilter mapsonly // Download only maps from server custom files
 //cl_downloadfilter none // Do not download anything from server custom files
 
+alias download_auto"alias download"
```
mastercoms: I would rather use the shortcut for auto than a full module level.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -613,10 +615,11 @@ alias sprays_on"cl_allowdownload 1;cl_allowupload 1;r_spray_lifetime 1;cl_player
 //violence_hblood 0 // Enable silly gibs/blood (less performance)
 //tf_playergib 1 // Control gibbing on local servers: 0: never, 1: default, 2: always.
 
+alias sillygibs_auto"alias sillygibs"
```
mastercoms: I would rather use the shortcut for auto than a full module level.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -761,10 +764,10 @@ alias textures_very_high"mat_filtertextures 1;mat_picmip -1;mat_disable_fancy_bl
 //rope_wind_dist 0 // Do not apply CPU intensive wind to ropes
 //rope_averagelight 1 // Only use average light, instead of an extra max intensity average with zero
 
-alias ropes_off"r_drawropes 0;rope_rendersolid 0;r_ropetranslucent 0;rope_smooth 0;rope_subdiv 0;rope_collide 0;rope_wind_dist 0"
-alias ropes_low"r_drawropes 1;rope_rendersolid 1;r_ropetranslucent 0;rope_smooth 0;rope_subdiv 1;rope_collide 0;rope_wind_dist 0"
-alias ropes_high"r_drawropes 1;rope_rendersolid 1;r_ropetranslucent 1;rope_smooth 1;rope_subdiv 2;rope_collide 0;rope_wind_dist 0"
-alias ropes_ultra"r_drawropes 1;rope_rendersolid 1;r_ropetranslucent 1;rope_smooth 1;rope_subdiv 7;rope_collide 1;rope_wind_dist 1000"
+alias ropes_off"r_drawropes 0;rope_rendersolid 0;r_ropetranslucent 0;rope_smooth 0;rope_subdiv 0;rope_collide 0;rope_wind_dist 0;rope_averagelight 0"
```
mastercoms: No benefit to setting 0. It has been intentionally removed from ropes.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1005,7 +1006,7 @@ tf_backpack_page_button_delay .25 // Decrease button delay for moving items
 //cl_crosshair_green 39 // Crosshair green (0 to 255)
 //cl_crosshair_blue 176 // Crosshair blue (0 to 255)
 //cl_crosshair_scale 32 // Size of crosshair
-//cl_crosshair_file"" // Style of crosshair
+//cl_crosshair_file "" // Style of crosshair
```
mastercoms: Space not needed

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1210,7 +1211,7 @@ alias outlines_high"glow_outline_effect_enable 1;tf_enable_glows_after_respawn 1
               // 0 - to console
               // 1 - to console and chat
               // 2 - to console and HUD
-//ds_prefix"" // Prefix file names with this text
+//ds_prefix "" // Prefix file names with this text
```
mastercoms: Space not needed

config/templates/modules/modules.cfg
```diff
@@ -116,7 +116,7 @@
 //3dsky=on
 // on, off
 
-//jigglebones=on
+//jigglebones=force_on
```
mastercoms: Force on should not be default here.

config/mastercomfig/cfg/comfig/finalize.cfg
```diff
@@ -1,3 +1,4 @@
 // Runs after user configs and addons
 mat_savechanges // Make sure we init next time with our material system cvars to prevent reloads
-block_antialias
+block_antialias // Block mat_antialias
+dynamic_background // Run dynamic_background after everything
```
mastercoms, line 4: What is the actual benefit to this? Seems like nothing.

mastercoms (changes requested):

.github/ISSUE_TEMPLATE/bug_report.md
Tiagoquix: The idea is to make that so the user does not have to delete the warning sign to say his comfig version. I think it would be better like this:

``* Config version<!-- don't say latest, say the actual version, found by running `version_comfig` in console -->:``

mastercoms: I am fine with that

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: The ideia, along with download=auto and sillygibs=auto is that, if the user wants to enforce the module like other modules (textures=high being textures_high, for example), he can have this in his dispose.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Alright, I'll revert this change.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Are you sure? I think it is. Since if you use `cl_crosshair_filecrosshair3` it won't work since it will interpret the whole thing as an one cvar.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Are you sure? I think it is. Since if you use `ds_prefixSOMETHING` it won't work since it will interpret the whole thing as an one cvar.

config/templates/modules/modules.cfg
Tiagoquix: The `modules.cfg` file should follow a pattern.

By my analysis, I concluded that the values that are set by default in the comments are the values that are the higher value the module can have (by looking at the docs page, from below to above), or if they have a universal default (a default on almost all presets), they become the comment's default value. For instance, `bandwith` is `1.5Mbps` and not `8.0Mbps` that is the max value the module offers. But, `lighting` is `ultra`, and it is the max value the module have.

And, with this conclusion, I changed the `jigglebones` default module level to `force_on`.

config/mastercomfig/cfg/comfig/finalize.cfg
Tiagoquix: Mainly to avoid some problem that may occur. Also, I think that is best to run a `map_background` after everything has been set.

.github/ISSUE_TEMPLATE/bug_report.md
Tiagoquix: Done.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Done.

config/mastercomfig/cfg/comfig/comfig.cfg
mastercoms: quotes are interpreted as separate arguments.

config/templates/modules/modules.cfg
mastercoms: Except I'm telling you what my logic was for the defaults, and this should be on :)

config/mastercomfig/cfg/comfig/comfig.cfg
mastercoms: I don't think that's really valuable, as that isn't the logic behind directly running a module level. Directly running a module level does not select a module, it applies the settings in that level. auto levels have no settings to apply, thus they would be blank.

config/mastercomfig/cfg/comfig/comfig.cfg
mastercoms: quotes are interpreted as separate arguments.

config/templates/modules/modules.cfg
Tiagoquix: Nice answer. You looked like a queen giving me orders now. I will revert this change.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Done.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Done.

config/templates/modules/modules.cfg
Tiagoquix: Done.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: I understood you point of view, but I don't think you understood mine.
For example, if an user uses `sillygibs_on` in the console, he only can revert the change to `sillygibs=auto` if him uses `sillygibs=auto;run_modules` since `sillygibs_auto` doesn't exist.

micwoj92: but like the point of auto is to not change what people set, so your argument here makes no sense.

mastercoms: Your implementation would also require a run_modules.

Tiagoquix: @micwoj92 "but like the point of auto is to not change what people set, so your argument here makes no sense."
The point of `auto` level is to let the config change the values, not to let the user change them. The only exception is the `download` module.

@mastercoms No, it won't.
For example: if a user has `snapshot_buffer=safe` in his `modules.cfg` and then he executes `snapshot_buffer_auto` into the console, it won't require a `run_modules`, since the function of `snapshot_buffer_auto` will act like the user is using `snapshot_buffer=auto`.

mastercoms: `snapshot_buffer=auto` previously had the same exact content as you have now for `snapshot_buffer_auto`, what are you talking about?

mastercoms: Disregarding what is the same, all you are doing here in your new changes is making `snapshot_buffer=auto` set the snapshot buffer module to set snapshot buffer to auto, which is just a horribly convoluted set up which doesn't act like any other module.

Tiagoquix: @mastercoms Yes but, following the config pattern, user's won't think that `snapshoft_buffer=auto` can be used the same way as `snapshoft_buffer_auto`. For example, you can use `lighting=ultra` but you will need `run_modules` for that, and by using `lighting_ultra` you don't need to run modules.

At least, might I alias `snapshot_buffer_auto` to `snapshot_buffer=auto` to it be like a shortcut?

mastercoms: **Your changes**

` alias snapshot_buffer_auto"alias snapshot_buffer" `

`snapshot_buffer_auto` sets `snapshot_buffer` to empty (auto state).

`alias snapshot_buffer=auto"alias snapshot_buffer snapshot_buffer_auto"`

`snapshot_buffer=auto` sets `snapshot_buffer` to `snapshot_buffer_auto`, which sets `snapshot_buffer` to auto state. Therefore, on the actual run of this module, you are only simply making it then clear itself on a second run. This does not work, and means the module level has two useless states depending on how many times modules are run.

mastercoms: **Original**

`alias snapshot_buffer=auto"alias snapshot_buffer"`

`snapshot_buffer=auto` sets `snapshot_buffer` to empty (auto state).

Tiagoquix: Alright, I'll revert these changes but I'll alias `snapshoft_buffer_auto` to `snapshot_buffer=auto` to be a shortcut.
Like I said, the config follows a pattern.

mastercoms: And finally, if you look at _any other module_, you will see that the actual level command never interacts with the module. It only applies the game settings. This is just how modules are organized.

For example, `packet_rate_big`, it sets a few cvars involving max routable, compression, etc. It does not do anything with the packet_rate module itself.

mastercoms: No, do not do this. It creates a module level which is unlike anything else, and creates states within the config which are either useless or harmful. `snapshot_buffer_auto` should _only_ apply the settings which are associated with an `auto` level, i.e. **nothing**. Your change **breaks** the config's pattern.

Tiagoquix: Done.

mastercoms (approved):

PR #400, opened 2020-10-07
Match HUD title smaller header

KKire:
Added missing # so Match HUD header is the same like the rest as it's part of HUD modules category


micwoj92:
big change, did you test on all supported platforms?

Tiagoquix:
Good observation.

mastercoms (approved):

PR #419, opened 2020-11-25
Config updates

Tiagoquix:
Changes:

- Removed `packet_rate_max`, since it's not document, there is no reason to be present in the config. I guess no one uses this. And if someone uses this, probably is no more than 10 people.
- Enabled light cache on `lighting_ultra`. I don't see a reason to it be disabled.
- Removed `water_medium_high` since it's pratically equal to `water_high` (the difference is that the entity reflects are enabled in high, while in medium-high they aren't).
- Removed `water_very_high` since it's also pratically equal to `water_ultra` (the difference is that `r_waterforceexpensive` is enabled in ultra, while in high it isn't).
- Removed `particles_medium`. The only difference between `particles_medium` and `particles_low` is `r_drawflecks` being enabled.
- Renamed `particles_low` to `particles_medium` and `particles_very_low` to `particles_low`.
- Reorganized the motion blur module and added some commands to `motion_blur_low`.
- Unified the anti-aliasing modules. Alpha coverage is now enabled when MSAA is enabled too.
- Increased `ragdoll_sleepaftertime` value in `ragdolls_medium` to `2.5`.
- Forced `mat_viewportupscale` to `1` since `mat_viewportscale` is also forced.
- Removed `mat_texture_reload_frame_swap_workaround` from `opengl_default` and setted it to `0` in `comfig.cfg`.
- Disabled forcing `tf_hud_target_id_alpha` to `100` by default.
- Reorganized the messages and debug modules.
- Forced `mat_levelflush` to be enabled by default.
- Enabled `mod_support_off` to be used again.

I also changed some values in the shadows module.

Note: the docs have not been changed.


mastercoms:
> Removed packet_rate_max, since it's not document, there is no reason to be present in the config. I guess no one uses this. And if someone uses this, probably is no more than 10 people.

It isn't documented since it isn't used in normal circumstances, but is useful for people who play on higher tickrate servers which do exist. I do want to keep this.

> Enabled light cache on lighting_ultra. I don't see a reason to it be disabled.

It reduces lighting quality because it needs to in order for the values to be cacheable in a reasonable amount of space, it does less accuracy of lighting.

> I also changed some values in the shadows module.

I don't see any reason why these values were changed.

> Removed water_medium_high since it's pratically equal to water_high (the difference is that the entity reflects are enabled in high, while in medium-high they aren't).
> Removed water_very_high since it's also pratically equal to water_ultra (the difference is that r_waterforceexpensive is enabled in ultra, while in high it isn't).

Just because it's one cvar that was changed doesn't make it practically equal. These are explicit quality levels within the game, so we should keep them too.

> Removed particles_medium. The only difference between particles_medium and particles_low is r_drawflecks being enabled.
> Renamed particles_low to particles_medium and particles_very_low to particles_low.

Again, this is enough for another module level. No need to do these sorts of cleanups when it is obviously intended for one cvar to change, as they exist in this fashion. Also, renaming very_low to low will cause tons of confusion.

> Reorganized the motion blur module and added some commands to motion_blur_low.

Falling intensity is zero, no need to add them to low. You moved the falling cvars away from falling intensity.

> Unified the anti-aliasing modules. Alpha coverage is now enabled when MSAA is enabled too.

Alpha to coverage is a much cheaper effect than MSAA. If you really want to combine them, I would make it anti_aliasing=on (for A2C) and then =msaa_2x, etc.

> Removed mat_texture_reload_frame_swap_workaround from opengl_default and setted it to 0 in comfig.cfg.

No need. It's default to 0 on Windows, and doesn't do anything anyway.

> Increased ragdoll_sleepaftertime value in ragdolls_medium to 2.5.

Why?

> Forced mat_viewportupscale to 1 since mat_viewportscale is also forced.

These are different things though.

> Forced mat_levelflush to be enabled by default.

No need, it already defaults to 1 and is not an archive or hardware config cvar.

> Enabled mod_support_off to be used again.

No, this was an intentional change.

Tiagoquix:
> It isn't documented since it isn't used in normal circumstances, but is useful for people who play on higher tickrate servers which do exist. I do want to keep this.

> It reduces lighting quality because it needs to in order for the values to be cacheable in a reasonable amount of space, it does less accuracy of lighting.

> Just because it's one cvar that was changed doesn't make it practically equal. These are explicit quality levels within the game, so we should keep them too.

> Again, this is enough for another module level. No need to do these sorts of cleanups when it is obviously intended for one cvar to change, as they exist in this fashion. Also, renaming very_low to low will cause tons of confusion.

> Falling intensity is zero, no need to add them to low. You moved the falling cvars away from falling intensity.

> No need, it already defaults to 1 and is not an archive or hardware config cvar.

All right. I'll revert these changes.

> I don't see any reason why these values were changed.

I think the values you chose aren't good. I'll increase the values even more.

> Alpha to coverage is a much cheaper effect than MSAA. If you really want to combine them, I would make it anti_aliasing=on (for A2C) and then =msaa_2x, etc.

I'll do this.

> No need. It's default to 0 on Windows, and doesn't do anything anyway.

It is defaulted to 0 on Linux and macOS too? If yes, we should stop forcing it to 0.

> Why?

Because cl_ragdoll_fade_time is 10 in ragdolls_high, and it is 5 on ragdolls_medium. Then, I did use the logic of the values in the medium level being the half of the values on the high level. That's why.

> These are different things though.

Should I keep this enforcement?

> No, this was an intentional change.

Why?

Tiagoquix:
I forgot to mention, but, I disabled cl_drawmonitors on effects_medium. I don't see any reason to it be enabled here, since the purpose of the medium level on the effects module is only to enable water splashes and first person bullet tracers.

Tiagoquix:
More updates:

The docs have been updated;
Misconceptions page on docs has been updated;
The transparent viewmodels addon CFG has been updated;
shadows_very_low has been restored.

I did not update the files inside `data`.

Tiagoquix:
I will do another pull request that will be with less changes (and only needed changes) because this has gone too far with unnecessary changes, making harder to analyze.

PR #422, opened 2020-12-12
Changed Power Plan Section

Sirbucket:
This explains that the ultimate performance plan is almost no different to high performance beyond pulling unnecessary amounts of power to test machines and that high performance uses too much power for a laptop battery to handle, reducing it's life span.


Sirbucket:
The video I linked does a good job at going over the power plans.

mastercoms:
I don't really agree with the conclusions here that result in these changes to the tweaks, sorry.

Sirbucket:
No problem

PR #425, opened 2020-12-23
Config updates

Tiagoquix:


Tiagoquix:
@mastercoms Sider ignores the "Cancelling" but doesn't approve the check. Can you fix that?

PR #426, opened 2020-12-28
Change how match_hud_off works to actually disable the match HUD

Tiagoquix:
What this does is: after `match_hud_once` has been ran, `match_hud_off` will be set to `tf_use_match_hud 0`, instead of keep being set to `tf_use_match_hud 1;alias match_hud_once match_hud_disable`.

**Yes, the module already works, but it doesn't work like all modules. For example, `textures_very_high` will set textures to very high instantly, but `match_hud_off` will only set the match HUD to be disabled in the first load.**

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -929,9 +929,10 @@ alias hud_panels_on"tf_healthicon_height_offset 10;tf_hud_target_id_offset 0"
 
 alias hud_panels hud_panels_on
 
-alias match_hud_disable"tf_use_match_hud 0"
+alias match_hud_alias"tf_use_match_hud 0"
+alias match_hud_alias2"alias match_hud_off match_hud_alias"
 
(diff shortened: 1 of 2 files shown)


mastercoms:
This is intentional, as setting it before a game will break the HUD. Also you have too many layers in your aliases, so the command doesn't actually run.

mastercoms:
Actually, I see what you're going for, so that subsequent runs of the module always set it off. What you can do instead is alias match_hud_off within match_hud_disable.

Tiagoquix:
> Actually, I see what you're going for, so that subsequent runs of the module always set it off. What you can do instead is alias match_hud_off within match_hud_disable.

No. I can't do that.
This way I did was to support `restore_config` (so running it won't reset match_hud_off).

mastercoms (approved):

mastercoms:
I see now, thanks!

Tiagoquix:
@mastercoms The problem can still happens if someone use run_modules and don't change classes. Should I resolve that?

Tiagoquix:
@mastercoms Resolved in #427

PR #427, opened 2020-12-28
match_hud_alias3

Tiagoquix:
This fixes a rare case when someone uses run_modules but doesn't change classes, so match_hud_alias2 will not be run again (since it is on game_overrides.cfg). But, with this change, match_hud_off will always disable the match HUD if match_hud_alias2 has been ran before.


config/mastercomfig/cfg/comfig/game_overrides.cfg
```diff
@@ -1,7 +1,7 @@
 game_overrides // Run aliased game overrides
 game_overrides_c // Run custom aliased game overrides
 game_overrides_once // Run commands that should be run once
-alias block_game_overrides_once"alias game_overrides_once";block_game_overrides_once // Block game_overrides_once
+alias block_game_overrides_once"alias game_overrides_once;match_hud_alias2";block_game_overrides_once // Block game_overrides_once and set match_hud_off to actually disable the match HUD
```
Tiagoquix: I don't know why this is being shown, but the comment is the same as the current comment.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -929,9 +929,10 @@ alias hud_panels_on"tf_healthicon_height_offset 10;tf_hud_target_id_offset 0"
 
 alias hud_panels hud_panels_on
 
-alias match_hud_disable"tf_use_match_hud 0"
+alias match_hud_alias"tf_use_match_hud 0"
```
Tiagoquix, line 932: This line is exactly equal compared to the current line present in the comfig.cfg file.

PR #429, opened 2020-12-29
Fix match_hud_off not working when run_modules has been used

Tiagoquix:
This fixes a rare case when someone uses run_modules but doesn't change classes, so `match_hud_alias2` will not be run again (since it is on `game_overrides.cfg`). But with this change, `match_hud_off` will always disable the match HUD if `match_hud_alias2` was ran before.


Tiagoquix:
Forget this. I had thought that `run_modules` would alias `match_hud_off`, I was confused. Actually, `run_modules` will only run `match_hud` which is already defined as `match_hud_off`, and `match_hud_off`, after spawning in a match, will already be defined as `match_hud_alias` (that is, it will work like other modules).

PR #430, opened 2020-12-31
Docs changes

Tiagoquix:
The changes that I made are just perfectionist changes. I just changed a text here and there.

Two changes I would like to highlight are:

- I removed the suggestions for disabling Game Mode and Full Screen Optimizations. Windows 10 has evolved over time and these changes are no longer strictly necessary. So much so that new games, such as Shadow of the Tomb Raider and Hitman 2 already feature a new full screen mode called Exclusive Full Screen, in which it was made exclusively to benefit from these Windows 10 changes.

- The removal of the "No lightwarps" lines from the medium and medium_high levels of the lighting module. These changes were made based on this commit: https://github.com/mastercomfig/mastercomfig/commit/f7ba3c450457ee7a018f0658bc226e9bc50412bd

PR #433, opened 2021-01-01
An addendum to #430

Tiagoquix:
This modifies the `modules.json` to better corresponds to the current modules page on the docs.
Also, I removed the `"tags": ["advanced"]` from `snapshot_buffer` because it was the only module that was using this, and I don't see a reason to keep it.

Please, only approve and merge this if #430 is merged.


mastercoms (changes requested):
The changes to advanced should be reverted, as well as the order of snapshot buffer.

Tiagoquix:
> The changes to advanced should be reverted, as well as the order of snapshot buffer.

I will revert the change related to the advanced, but I won't revert the order of snapshot buffer. I am [just listing it the way it's listed in the docs](https://docs.mastercomfig.com/en/latest/customization/modules/#snapshot-buffer).

Tiagoquix:
Done.

mastercoms:
The docs should be changed if you insist.

Tiagoquix:
Done @mastercoms

mastercoms (approved):

PR #434, opened 2021-01-01
Removed command aliases

PazerOP:
On a local server, these aliases cause any invocation of these commands result in a feedback loop, where every tick there are n + 1 invocations of that command being processed. This quickly causes the client to disconnect from the local server with "issued too many commands to server". If you increase the maximum number of commands per second before that check is hit, you will realize that all of these command are also broken on local servers (they are never actually processed). You just have a forever-increasing queue of commands building up, and if you look in net graph you will see an ever-increasing amount of network traffic.

ConCommands are stored in a hash map, which in practical terms means that searching through them happens in constant time, or O(1). On average, the time to look up any given command by name is going to be the same as any other command. So here, instead of spending your one unit of time looking up `explode`, for example, you spend your one unit of time looking up `cmd`. All this does is break local servers.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -31,33 +31,6 @@ hud_escort_interp .1 // Smooth out payload cart progress over slightly less time
 net_chokeloop 1 // Early check for multiplayer
 sv_lan 1 // Protect local servers
 
-// Directly forward string cmds to server for lower latency
-alias resetclass"cmd resetclass"
-alias menuopen"cmd menuopen"
-alias menuclosed"cmd menuclosed"
-alias stop_taunt"cmd stop_taunt"
-alias td_buyback"cmd td_buyback"
-alias arena_changeclass"cmd arena_changeclass"
-alias extendfreeze"cmd extendfreeze"
-alias show_motd"cmd show_motd"
-alias showroundinfo"cmd showroundinfo"
-alias autoteam"cmd autoteam"
-alias boo"cmd boo"
-alias done_viewing_loot"cmd done_viewing_loot"
-alias spectate"cmd spectate"
-alias status"cmd status"
-alias ping"cmd ping"
-alias pause"cmd pause"
-alias unpause"cmd unpause"
-alias demorestart"cmd demorestart"
-alias fade"cmd fade"
-alias nextmap"cmd nextmap"
-alias timeleft"cmd timeleft"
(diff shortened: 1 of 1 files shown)


Tiagoquix:
I can confirm this. Happened to me as well.

ClusterConsultant (approved):

Tiagoquix (approved):

mastercoms (changes requested):
> ConCommands are stored in a hash map

Where did you find that to be the case? Commands and console variables are a linked list, and `FindCommand(const char* name)` loops through this linked list, doing a string comparison, which is `O(n^2)`, one for the number of commands, and another for the length of the string.

Since these commands do not exist on the client DLL, the client must loop through all client commands and console variables before realizing it is an unknown command on its own end, and then sends the server a string command in hopes that it may be found. So for every one of these commands, the client unnecessarily loops through all of its own commands and cvars, for each of its matching characters, or a minimum of the first character.

Now that we've established this to be the case, let's actually look at what the problem is here, since this PR at its current state is a bit imprecise. The problem is that some of these commands already have this logic built-in, and define themselves on the client/engine, to manually forward to the server in the case of being executed on the client. I didn't think there would be any harm in aliasing these, but evidently, I was wrong here. These commands are: `status`, `pause`, `unpause`, and `ping`. No other commands are broken on local servers (I just tested each one to confirm my analysis of the source code). I would be fine with removing these from the config, as they don't really impact play on multiplayer servers and do break local servers.

So, please change the PR such that it only removes these commands from the forwarding aliases.

PazerOP:
In src2007, they were indeed a linked list. As of 2017, they exist in both data structures, and FindCommandBase uses the hash map. I haven't confirmed if those changes exist in tf2 though. Even if they weren't, there are so many other commands that are being executed constantly, it doesn't make sense to try to micro-optimize these commands that are run so infrequently. 

> as they don't really impact play on multiplayer servers

In what universe is *command lookup* the bottleneck when you are running `demorestart`?

mastercoms:
I don't think they do in TF2, and I have measured a difference between the two. There's many cases where it's simply more pleasant for the game to be slightly more responsive to the user, and mastercomfig these days is all about micro-optimizations. If you could point towards where you saw this, it would be helpful, because I haven't seen an indication of this in the leak or latest symbols.

> In what universe is command lookup the bottleneck when you are running demorestart

When you are spamming it every few milliseconds to force full updates, at the cost of greater network usage.

mastercoms (approved):

PR #435, opened 2021-01-05
Adjustments for lighting and shadow modules

Tiagoquix:
`lighting` : Max dynamic lights has been increased to 8 and 16 on high and very high, respectively.

`shadows` : I turned all values ​​into even values ​​and restored nb_shadow_dist in shadows_low.


mastercoms:
I like the values how they are after my own tuning.

PR #436, opened 2021-01-05
General config updates

Tiagoquix:
List of changes:

- The majority of the changes are related to comments. I adjusted some comments to follow the contributing guidelines. Also, some aliases have also been adjusted to follow the pattern of "which command appears first will be listed first". For instance: `mat_hdr_level` is the first commented command in the post-processing section, so it is the first command that appears on the modules' levels. I also restored some quotes/spaces in some scripts' files.
- `debug=on_partial` has been removed. I don't think anyone uses it.
- I aliased some commands to be with an full alias since they have spaces on them. This is to guarantee that they work properly. For instance: `cmd VModEnable 1` is now `voice_chat_on_alias`, and I also did this to the SourceTV commands.
- Decals have been increased to 96 in `decals_high`. `r_decal_cover_count` has been increased to `4` on `decals_medium` and `decals_high` (this was made to follow the patern from the other levels).
- I changed `modules_run.cfg`, because silly gibs is already ran by the gibs module and I also changed the mod_support position to align with the docs.
- I changed a comment regarding the Medi Gun in the Very Low preset.
- Some commands were commented out to stop enforcing their default values (since there is no need to).
- I changed some values in the `motion_blur` and in the `props` modules.
- `tf_chat_popup_hold_time` has been increased to `4.5` (from `3`).
- `r_worldlightmin` has been changed to `.0008` from `.002` on `lighting_low`.
- "Misc modules" have been renamed to "Miscellaneous modules" in the docs.

That's all.


mastercoms:
I've ported over all the changes which I agree with.

PR #440, opened 2021-01-08
TF2 update made these invalid

KaineWooten:
Due to the update that now force sets your cmdrate and updaterate to 66 I don't feel there is a reason to really keep these as options

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -79,11 +79,6 @@ alias interp_standard_safe"cl_interp .030303;setinfo cl_interp .030303"
 alias interp_standard_low"cl_interp .015151;setinfo cl_interp .015151"
 alias interp_standard_spy"cl_interp .022499;setinfo cl_interp .022499"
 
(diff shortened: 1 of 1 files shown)


micwoj92:
There are still custom servers where it should work, right?

KaineWooten:
Why would you want to include custom servers when this config is usually used for Casual servers and Comp (RGL/UGC/ETF2L) servers. Which by default will have it set to 66, custom servers are irrelevant because the ones that would allow these commands to go higher are in the very low minority

micwoj92:
This config is intended for all TF2 players. Also funny that you mention comp as most servers with 100 tickrate are mge servers.

KaineWooten:
MGE is DM for 1 on 1 deathmatches. 

ALL TF2 players yes but these settings aren't going to be utilized at all whilst in a server. A vast majority of servers to be more precise

mastercoms:
There was never a bug which allowed players to use a higher update or cmd rate. There was a bug that allowed players to constrain their interp by an unbounded update rate. This module level was never related to this. It was explicitly added for high tickrate servers.

PR #447, opened 2021-01-29
Filter junk strings out of the launch option lists

WhyIsEvery4thYearAlwaysBad:
I made this change because with the TF2 code leak, we can get a better reference to actual launch options that exist in the currrent game, and filter out any bad launch options (strings that are not actually launch options) from the current launch option list. This way the user isn't misinformed on what are actually launch options, and can have some reliable information to work off of. Keep in mind that this list is *not* final yet, since I cannot confirm what launch options in the linux list are actually in the game. I can do the same on the Windows list, but I have not done so yet.

P.S A lot of launch options here seem to be only for MacOS (marked in block/inline-style comments). Should I move them to a MacOS launch option list or just keep them in the linux launch op list?


treacherousfiend:
I seem to remember being told that this launch option list should stay up to date at all times, just in case they add or remove a launch option, which is why its specifically extracted strings from the exe file

WhyIsEvery4thYearAlwaysBad:
> 
> 
> I seem to remember being told that this launch option list should stay up to date at all times, just in case they add or remove a launch option, which is why its specifically extracted strings from the exe file

The point of this change is to reduce (or remove if possible) the amount of junk strings in both Windows and Linux launch option lists. I doubt the TF2 team would add/remove launch options fast enough (6 months to 1 year) for this to become an issue, but if it ever does I could always use the leaked launch options from the 2020 leak as a *filter* against the junk strings, and then insert the launch options that do pass the filter to get around that issue. (EDIT: Actually this is what I'm going to do instead, since that evades the risk of listing launch options that no longer exist.)

treacherousfiend:
Personally, I agree, but I tried to do the same thing and got shot down.

WhyIsEvery4thYearAlwaysBad:
Wait I do have an iMac, so I can get the MacOS version of the TF2 binaries, just not the Linux versions.

mastercoms (changes requested):
I won't accept changes directly referring to the code leak.

Also, the list of launch options should reflect the current state of TF2, and should be reproduced in an automated fashion, as the current contributing instructions suggest. Rather than completely replacing the launch options list, there ought to be a handmade list of blocked strings that are verified to not be launch options, which can filter out from the tool.

I think we had a conversation about both of these topics on TFTV, and it seems like you plan on carrying that out in your later comment that you posted here.

WhyIsEvery4thYearAlwaysBad:
> Rather than completely replacing the launch options list, there ought to be a handmade list of blocked strings that are verified to not be launch options, which can filter out from the tool.
> 
> I think we had a conversation about both of these topics on TFTV, and it seems like you plan on carrying that out in your later comment that you posted here.

Yeah this is exactly what I'm going to do. I just need to get my hands on the Linux TF2 binaries.

WhyIsEvery4thYearAlwaysBad:
How should I filter out the bad launch options out though? I was thinking to

WhyIsEvery4thYearAlwaysBad:
I've managed to generate a junk string blacklist, but haven't used it yet. Your launch option lists seems to be generated from client.dll and server.dll. Should I bother with the other binaries?

EDIT: I'm going to go ahead and use all of the TF2 binaries that are used in TF2 itself.

WhyIsEvery4thYearAlwaysBad:
After 5 hours I finally got the Linux binaries. I'll allow this pull request to be merged/reviewed after I filter out the linux launch op list.

WhyIsEvery4thYearAlwaysBad (commented):
@mastercoms I have changed the launch option lists. The launch options are no longer ripped directly through the 2020 TF2 leak, but instead those launch options are used as a whitelist against the junk strings to form a blacklist of junk strings, which then can be used to get most of the good launch options that are in the game.

mastercoms:
There's tons of launch options which are valid, that you have removed. I'll base my own work on this though, thank you.

PR #448, opened 2021-02-01
Reworked mm_override.cfg

Tiagoquix:
Based on the pull request #188, this is the new reworked `mm_override.cfg`.

Instead of looking at more than [3000 commands](https://gist.github.com/mastercoms/c4b65d588a486cd58c765eccce284144), I just did look at the `comfig.cfg` file and added the default values to `mm_override.cfg` (to, well, override the `comfig.cfg`'s values).

Please, take a deep look at the changes, because I'm not sure if all the commands that I did put are necessary.

That's all for now.


Tiagoquix:
@mastercoms Since some commands have to be at their default values so that the user can play the MM without crashes, the `mp_decals` command should be at its default value (200) or should it be at the value of 2048 (that is the default value of `r_decals`)? Because `mp_decals` should be the same value of `r_decals`.

mastercoms:
I don't think this change is necessary, as every value does not need to be reverted. We only need to change the ones which cause crashes when they are changed in the loading screen. And we have already done that.

PR #449, opened 2021-02-05
Add Medium-Low gibs module

treacherousfiend:
As pointed out by someone in our discord, the low module for gibs limits them to 2 pieces per object, and medium doesn't limit them at all. A quick check ingame shows that an exploding player can spawn anywhere between 9-12 gibs based on the class.

This Medium-Low module attempts to create a sweet spot between the two modules by limiting gibs to 4 per object.

The value of 4 gibs per object has little basis, and other values may fit better.

PR includes new module and it's addition to the docs.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -615,6 +615,7 @@ alias sillygibs
 
 alias gibs_off"cl_burninggibs 0;props_break_max_pieces 0;tf_playergib 0;sillygibs_on;sillygibs"
 alias gibs_low"cl_burninggibs 0;props_break_max_pieces 2;tf_playergib 1;sillygibs_off;sillygibs"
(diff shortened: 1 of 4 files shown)


micwoj92:
also needs changes data/modules.json

Tiagoquix:
Hello. Please remove the spaces from ``alias gibs_medium_low"cl_burninggibs 0; props_break_max_pieces 4; tf_playergib 1;sillygibs_off;sillygibs"``, and also, update your [modules.cfg file](https://github.com/treacherousfiend/mastercomfig/blob/medium_low_gibs/config/templates/modules/modules.cfg) (the template file).

Remember that your change will not force this new level of the gibs module to be automatically enabled for everyone in some preset. You would have to change some preset (like the medium-high), but I do not suggest you to. This is just a reminder.

mastercoms (changes requested):
Add to modules template

treacherousfiend:
fixed

mastercoms (approved):

PR #452, opened 2021-02-07
Don't force  party join request mode

micwoj92:
Don't force  party join request mode

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -183,7 +183,7 @@ alias download
 tf_mm_debug_level 0 // Remove matchmaking debug output
 //tf_mm_partyclient_debug 1 // Enable party debug output
 //tf_party_ignore_invites 1 // Ignore party invites
-tf_party_join_request_mode 1 // Mode for party join requests: 0 - open join, 1 - request join, 2 - invite join.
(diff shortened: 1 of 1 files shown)


Tiagoquix:
I do not agree with this change. It is annoying that anyone could join your party without asking if they could join in the first place.
CS:GO's default, if I'm not mistaken, is a "ask-first-permission" to join someone's party.

micwoj92:
Mastercomfig should not change settings which do not matter for performance. You can check that many changes in recent months were like this. I will repeat what I said in #360 "Keeping in line with the new approach to mastercomfig defaults, the default for the downloads module should not be set." Just replace download module with party join request mode.

micwoj92:
image
This is what mastercoms said about this

micwoj92:
Also please do not comment on my prs unless you unblock me so I can comment on yours, thanks.

Tiagoquix:
Your reason for making this change is valid, but I still don't agree with the change.

mastercoms:
This is a change made for security reasons as someone who joins your party can spam party chat.

melvyn2:
I find this quite annoying, as I have to change it at launch every time, at least until I found the command and the cause. It would be better for it to be marked somewhere that this option is set by default, and have an option to disable it.

mastercoms:
I would be open to a module for this option.

Tiagoquix:
> I find this quite annoying, as I have to change it at launch every time, at least until I found the command and the cause. It would be better for it to be marked somewhere that this option is set by default, and have an option to disable it.

Fixed in #455

PR #471, opened 2021-03-31
Remove unused holiday punch texture

treacherousfiend:
also removed a couple crates. could probably remove mvm loot and taunts but i see no reason to rn.


treacherousfiend:
i do want to mention that i have no clue what the heavy christmas lights (xms_coloredlights) are used for. i couldn't find any cosmetics or weapons that use it but i could be wrong.
Engi guitar and Pyro "bubble wand" are used for taunts despite not being in the taunt folder.
i'm about 99% sure the "knife shield" for Sniper is the only texture for the razorback because i couldn't find any others. it's just in the Sniper folder and not the weapon folder
No clue what "scout_skeleton" is either, i didn't see a cosmetic using it and the Bat outta hell uses different textures

treacherousfiend:
revised PR to only remove the unused holiday punch texture at the request of coms

mastercoms (approved):

mastercoms:
Thanks for your contribution to mastercomfig!

PR #475, opened 2021-04-01
Update mm_override.cfg

Tiagoquix:
Added:
- `cl_showfps 0`
- `muzzleflash_light 1`
- `phonemedelay 0`
- `r_drawtracers_firstperson 1`
- `props_break_max_pieces -1`

Removed various commands that the config doesn't touch anymore (like r_threaded and r_queued commands).

Corrected the default values for:
- `mp_decals` : 2048 -> 200
- `r_flashlightdepthres` : 512 -> 1024

`ragdoll_sleepaftertime` is now `5` instead of `5.0f`, and lastly, the file is now sorted alphabetically.


Sirbucket:
From what I know this looks good seeing as multithreading is pretty trash in TF2

mastercoms:
Personally, I think `mm_override.cfg` should be stripped down to only commands which potentially cause crashes during loading into competitive.

PR #489, opened 2021-05-18
Add decals_medium_high

Tiagoquix:


mastercoms:
A change between 64 and 80 is really not a huge difference, so I don't think this module adds much value. 80 was selected for the decal LRU params, so I do not want to change it.

mastercoms:
The rest of the changes in this PR don't seem valuable either.

PR #492, opened 2021-05-22
#489 without decals changes

Tiagoquix:
Justifications for each change:

**comfig.cfg**

- `mat_powersavingsmode`: the comment isn't explaining what it really does. The current comment was added by me when mat_powersavingsmode was in the Optimizations section. It had the same idea as the `mat_vsync` comment (`mat_vsync 0 // Ensure VSync is disabled`).

- `dynamic_background`: I added `sv_allow_wait_command 1` to it because I thought that in the future someone would say "hey my map background isn't loading whatdoido helpmepls" so here is the preventive solution.

- `debug_on_partial`: I changed it so it only enables the developer console, net_graph 1 and cl_showpos 1.

- The rest of the changes are purely cosmetic. Most of them were made to align with the docs.

**modules.md**

- Changed the explanation for `debug=on_partial`.

**troubleshoot.md**

- The OpenGL errors don't appear only a few times when using Windows, they always appear in the console after loading the game, so I removed this part of the troubleshoot.

That's all.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -800,7 +800,7 @@ mat_managedtextures 1 // Force enable managed textures
 // ---------------
 // Controls the highest frame rate (FPS/frames per second) that the game can reach.
 
-//mat_powersavingsmode 0 // Ensure power savings mode is disabled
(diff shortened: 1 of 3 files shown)


KaineWooten:
Couldn't you have let `sv_allow_wait_command 1` to be its own independent line? Just seems unnecessary to only have it incorporated into the dynamic background alias

PR #493, opened 2021-05-26
Added VRR to windows.md

Sirbucket:
Variable refresh rate is now a baseline feature if your GPU supports it, you can find it in the same menu as Hardware-accelerated GPU scheduling.

### DIFF
--- a/docs/os/windows.md
+++ b/docs/os/windows.md
@@ -15,7 +15,7 @@ If you experience problems or reduced performance from this feature, turn it off
 
 ## Enable Variable refresh rate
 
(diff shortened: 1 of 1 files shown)


Sirbucket:
I noticed this in my graphics menu after the most recent creators update, so I put it here.

Sirbucket:
For confirmation 
VRR
https://www.techspot.com/news/80369-microsoft-adds-variable-refresh-rate-setting-windows-10.html

Sirbucket:
The article is slightly outdated as the current iteration works for all game platforms unlike the version initially released which only supported MS store games.

mastercoms:
Are you sure this works with the DX9 frame submission that TF2 uses? (i.e. no DXGI, no multiple swapchains)

mastercoms:
I am going to close this, as it will not work with TF2.

> This new toggle enables VRR support for DX11 full-screen games that did not support VRR natively, so these games can now benefit from your VRR hardware.

https://devblogs.microsoft.com/directx/os-variable-refresh-rate/

PR #499, opened 2021-06-07
Remove mp_respawnwavetime from debug

Tiagoquix:
It does nothing and `spec_freeze_traveltime`'s minimum value is `0.01`.


mastercoms:
Yes it does, it is still used in some scenarios. spec_freeze_traveltime will get clamped to the min value. No point in adding more decimals than needed.

PR #506, opened 2021-06-10
Config updates [WIP]

Tiagoquix:
The main purpose of this PR is to make possible to execute the module itself when you run its level. For example, `lod_ultra` doesn't change `lod` to `alias lod lod_ultra`. I am currently very confused on how to do it without breaking anything. `sillygibs` is a big obstacle to implement this.


Tiagoquix:
Forget this. This PR would break the concept of temporary module changing. From the docs:

> Since modules aliases act like normal commands, they can also be used in console to easily change game settings. So you could enter textures_high into console to temporarily set your texture quality to high for that game session. You can reset your modules back to what they were on startup by entering run_modules into the console.

PR #510, opened 2021-06-13
Changed debug_output to toggle developer cvar

TheDIGS:
Tiny PR but it's my first so fuck it.

All this changes is the `debug_output` command which currently sets `developer` to 1.

This changes that to toggle it instead, so running `debug_output` twice would toggle it from 0 to 1 to 0 etc.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -1174,7 +1174,12 @@ alias switchconsole console_on_flip
 // -------------
 // Some handy commands for debugging the game.
 
-alias debug_output"incrementvar developer 0 1 1"
+alias debug_output"debug_output_1"
+alias debug_output_toggle"incrementvar developer 0 4 1"
+alias debug_output_1"developer 1"
+alias debug_output_2"developer 2"
+alias debug_output_3"developer 3"
+alias debug_output_4"developer 4"
 
 alias debug_instant_respawn"sv_cheats 1;mp_disable_respawn_times 1;spec_freeze_time 0;spec_freeze_traveltime 0;mp_respawnwavetime 0"
 alias debug_invulernable"sv_cheats 1;buddha"
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -1174,7 +1174,7 @@ alias switchconsole console_on_flip
 // -------------
 // Some handy commands for debugging the game.
 
-alias debug_output"developer 1"
+alias debug_output"incrementvar developer 0 1 1"
 
 alias debug_instant_respawn"sv_cheats 1;mp_disable_respawn_times 1;spec_freeze_time 0;spec_freeze_traveltime 0;mp_respawnwavetime 0"
 alias debug_invulernable"sv_cheats 1;buddha"


micwoj92:
"Tiny PR but it's my first so fuck it."

You don't know what a tiny pr is, bro.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1174,7 +1174,7 @@ alias switchconsole console_on_flip
 // -------------
 // Some handy commands for debugging the game.
 
-alias debug_output"developer 1"
+alias debug_output"incrementvar developer 0 1 1"
```
mastercoms: This can use `toggle developer` instead. But we can also use `incrementvar developer 0 4 1` since that will cycle through all available levels.

mastercoms (changes requested):

config/mastercomfig/cfg/comfig/comfig.cfg
TheDIGS: are there more levels than 0-1?
Also, from what I've noticed, every time I've bound something in config.cfg to `toggle x`, Valve's translated it to `incrementvar x 0 1 1`. I have a hunch they're the same thing, but toggle is more abstracted.

mastercoms: Yes, it goes up to 4.

TheDIGS: Can I edit the PR or something? I'm kinda new to GH.

micwoj92: Yes, edit your branch

TheDIGS: Ok gonna do this in a minute

TheDIGS:
There are now multiple commands:
`debug_output` = `developer 1`
`debug_output_toggle` = run through all modes
`debug_output_<x>` = specific mode

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1174,7 +1174,12 @@ alias switchconsole console_on_flip
 // -------------
 // Some handy commands for debugging the game.
 
-alias debug_output"developer 1"
+alias debug_output"debug_output_1"
+alias debug_output_toggle"incrementvar developer 0 4 1"
```
Tiagoquix, line 1178: Only `debug_output"incrementvar developer 0 4 1"` would be sufficient.
The other aliases are kind of unnecessary.

Tiagoquix (changes requested):
See the comment below.

config/mastercomfig/cfg/comfig/comfig.cfg
TheDIGS: I don't think many people really want 3 and 4 anyway, so my reasoning is that more choice is better than more simplicity.

Tiagoquix: I disagree. I believe very few people actually use debug commands, so I don't think we should have that many commands. Anyway, it's up to you to decide on the final model and mastercoms will either reject or approve, but I prefer fewer commands.

mastercoms: I am fine with digs' solution.

mastercoms (approved):

Tiagoquix:
The docs should be changed too @mastercoms @TheDIGS

treacherousfiend:
Although this PR has already been merged, I think we should do some more research. The wiki states that there are 4 values for `developer`, but i can't get any more or less info to show up on `developer` 3 and 4. I also couldn't find references in the code to 3 or 4, but i might've missed it.

It's also worth pointing out that although its not mentioned on the wiki, TF2 (and potentially other source games) has `developer -1`, which outputs to the top left corner without adding anymore to the actual console output.

mastercoms:
Yeah that's right, 3 and 4 are unused in the current build.

PR #525, opened 2021-06-26
Update mkdocs.yml

Tiagoquix:


mastercoms:
Language name is not valid currently

Tiagoquix:
> Language name is not valid currently

Please revert the "(Brasil)" change then. Although Brazilian Portuguese is the most spoken of all portuguese languanges, it is not the unique.

PR #550, opened 2021-07-05
Modules - russian translation

Maurvick:


Maurvick:
This is not the best translation, and it still needs to be completed.

mastercoms:
The reason it says "found reference to undefined definition" is because you have a space in between your links.

It should be `[like](this)` rather than `[like] (this)`.

PR #563, opened 2021-07-20
Big config updates

Tiagoquix:
* Added a new module: hud_avatars - controls target ID avatars
* Added a new module: romevision - controls RomeVision off/on
* hud_panels now also controls target ID translucency
* Cleared lighting_ex in comfig.cfg to make it compatible with the None preset
* Fixed sheens_speed_level, sheens_tint_level and lighting_ex_level not working
* Various other changes


mastercoms (changes requested):
Don't change FPS max default. CS:GO has a different frame limiter mechanism which is not as problematic as TF2's. The higher bandwidth also had a lot of problems when it was set to 1.5Mbps.

Tiagoquix:
Done @mastercoms

mastercoms (approved):

PR #565, opened 2021-07-22
translation added

h9lk:
Will add one file per translation this topic to understand current state of affairs


h9lk:
don't understand~ how to deal with it, let it be so, fixed all mistakes in general :shipit:

PR #569, opened 2021-08-06
Add Scottish Resistance sticky outline fix

treacherousfiend:
Its a relatively common question, and considering it can be fixed with modules I feel like it should be mentioned either here or on the modules page.

### DIFF
--- a/docs/next_steps/quick_fixes.md
+++ b/docs/next_steps/quick_fixes.md
@@ -132,4 +132,4 @@ One other thing to note is that client interp is determined by whichever is the 
 
 ## Scottish Resistance stickybombs have no outline
 
-On some Mastercomfig presets, outlines are disabled. The outline settings include player and objective outlines as well as outlines for Scottish Resistance stickybombs. You can enable outlines by adding `outlines=medium` to your `modules.cfg`.
+On some mastercomfig presets, outlines are disabled. The outline settings include player and objective outlines as well as outlines for Scottish Resistance stickybombs. You can enable outlines by adding `outlines=medium` (or higher) to your `modules.cfg`.
--- a/docs/next_steps/quick_fixes.md
+++ b/docs/next_steps/quick_fixes.md
@@ -132,4 +132,4 @@ One other thing to note is that client interp is determined by whichever is the 
 
 ## Scottish Resistance stickybombs have no outline
 
-On some Mastercomfig presets, outlines are disabled. While the outline settings include player and objective outlines, it also includes the outlines for Scottish Resistance stickyboms. You can enable outlines by adding `outlines=medium` to your `modules.cfg`.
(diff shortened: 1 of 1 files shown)


Tiagoquix:
Does this fix work for DX8 users?

treacherousfiend:
I had no clue there were even issues with the SR on DX8, i'll have to do some testing

docs/next_steps/quick_fixes.md
```diff
@@ -129,3 +129,7 @@ cl_interp x
 ```
 
 One other thing to note is that client interp is determined by whichever is the greater of cl_interp_ratio divided by cl_updaterate, and the value of cl_interp. cl_interp_ratio and cl_interp have no interaction beyond that.
+
+## Scottish Resistance stickybombs have no outline
+
+Due to a bug in the game, the outlines for Scottish Resistance stickybombs only appears when objective outlines are enabled. You can fix this by adding `outlines=medium` to your `modules.cfg`.
```
mastercoms: It's not a bug, `glow_outline_effect_enable` controls all outline effects. I would just say that SR stickies require outlines, which can be enabled with at least outlines=medium module level.

mastercoms (changes requested):

docs/next_steps/quick_fixes.md
treacherousfiend: Should be better now? i'll let you resolve if you think its good.
I might've overdid it though, i'm not great at this sort of thing.

Tiagoquix: So this won't work on DirectX 8.

@treacherousfiend Please add a note about this.

docs/next_steps/quick_fixes.md
```diff
@@ -129,3 +129,7 @@ cl_interp x
 ```
 
 One other thing to note is that client interp is determined by whichever is the greater of cl_interp_ratio divided by cl_updaterate, and the value of cl_interp. cl_interp_ratio and cl_interp have no interaction beyond that.
+
+## Scottish Resistance stickybombs have no outline
+
+On some Mastercomfig presets, outlines are disabled. The outline settings include player and objective outlines as well as outlines for Scottish Resistance stickybombs. You can enable outlines by adding `outlines=medium` to your `modules.cfg`.
```
mastercoms:
```suggestion
On some mastercomfig presets, outlines are disabled. The outline settings include player and objective outlines as well as outlines for Scottish Resistance stickybombs. You can enable outlines by adding `outlines=medium` (or higher) to your `modules.cfg`.
```

mastercoms (changes requested):

mastercoms (dismissed):

docs/next_steps/quick_fixes.md
mastercoms: @treacherousfiend Please add a note about DX8 (in a new PR)

PR #572, opened 2021-08-14
ADD NOTE

micwoj92:


Tiagoquix:
I think this is unnecessary.

micwoj92:
I THINK THAT FLASHLIGHT MODULE IS UNNECESSARY.

mastercoms (approved):

PR #589, opened 2021-09-18
Fix portuguese translation accents

Tiagoquix:
Unfortunately there are several files, but at least the translation works correctly this way.


mastercoms:
I would rather remove translations than have this be the fix. We need to find another solution.

Tiagoquix:
> I would rather remove translations than have this be the fix. We need to find another solution.

I agree to remove the translations. Like, translating the docs is ok but translating the in-game echoes are unnecessary.

PR #590, opened 2021-09-18
Remove translation support

Tiagoquix:
The config "survived" more than 4 years without any translation. Besides translations not being that necessary, users would probably not use this module. And also, all console output is in English.


mastercoms (approved):

WhyIsEvery4thYearAlwaysBad:
Arguably people could use site translators like Google Translate or Yandex.Translate, but they are not perfect which I suppose is why mastercoms added translation support.

PR #598, opened 2021-09-30
Added OS specific recommended launch option settings.

Sirbucket:
Fits the design of the rest of the page and makes it easier for people to specifically grab what they want. This code should work, I'm not certain how to test it however.

### DIFF
--- a/docs/customization/launch_options.md
+++ b/docs/customization/launch_options.md
@@ -7,8 +7,22 @@ description: Tune Team Fortress 2 and mastercomfig further with launch options.
 Right click Team Fortress 2 in your Steam library and click **Properties...**. In the window that appears, under **Launch Options**, remove
 anything that's in there and paste
 
+=== "Windows"
+
 ```c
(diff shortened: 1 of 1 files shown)


Sirbucket:
Reasoning is just that people like copying and pasting things without reading why they're important and they may just input launch options that don't do anything/may be negative, however it's more or less just for convenience on the users end.

Tiagoquix:
Thanks for your suggestion, but it's kind of unnecessary. The page with the current format is great and meets all the needs of most users.

"Oh, but people don't read and don't even know what they're doing!" -- No problem at all, as the recommended launch options are for all operating systems.

Sirbucket:
> Thanks for your suggestion, but it's kind of unnecessary. The page with the current format is great and meets all the needs of most users.
> 
> "Oh, but people don't read and don't even know what they're doing!" -- No problem at all, as the recommended launch options are for all operating systems.

Mainly meant it cause -noquicktime does nothing on linux so maybe I'll bold it instead.

PR #599, opened 2021-10-01
Bolded critical parts of console commands plus grammar fix.

Sirbucket:
Well the last change was heavily unnecessary but I think this might at least get a bigger visual look to people that glance without it requiring more buttons to be pressed. Plus the current grammar looks like it was written in a rush, or at least to me as a native English speaker it looks wrong. It is missing punctuation, capitalization and feels too shortened in an unnecessary way.

### DIFF
--- a/docs/customization/launch_options.md
+++ b/docs/customization/launch_options.md
@@ -20,7 +20,7 @@ Read below about optional launch options and choosing your own DXLevel.
 === "Windows"
 
     !!! info
-        Add these launch options to your first launch, and then remove them.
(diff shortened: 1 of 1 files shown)


Sirbucket (commented):
More or less I feel like this should've happened a very long time ago either way and it's kind of strange to see how long it took for a grammar freak like me to find this.

Tiagoquix:
image

https://docs.mastercomfig.com/9.6.3/contributing/#launch-options

Sirbucket:
Really does not seem like proper formatting in the slightest but I won't judge.

PR #618, opened 2022-03-14
Config updates

Tiagoquix:
This pull request:

- **fixes `bandwidth=512Kbps` and `bandwidth=762Kbps` not working after the latest release;**
- makes some changes to `contributing.md`;
- removes various random spaces in `comfig.cfg`;
- replaces `net_splitpacket_maxrate`'s comment in `comfig.cfg` to a generic one, since the default rate is 80000 and the current value is not lower than that;
- changes `cl_hud_killstreak_display_alpha` to `120` (its default value) in `comfig.cfg`, like my other pull request;
- adds `hud_contracts=auto` which behaves in the same way as `download=auto`, rather than just leaving it as "unset";
- adds a warning to `sheens_tint` saying it does not work with `sheens_speed=off`;
- changes `config_template.cfg` to not disable HUD achievement tracking by default;
- updates `config_template.cfg` according to changes in `comfig.cfg`.

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -163,10 +163,10 @@ the existing modules if applicable, then add documentation for it at the
 
 The first part of adding modules is a multi-step process in `config/mastercomfig/cfg/comfig/comfig.cfg`:
 
-* Add the module level alias(es) (`alias module_level "cvar1 1; cvar2 0`)
+* Add the module level alias(es) (`alias module_level "cvar1 1;cvar2 0`).
   * For every command in the module, all levels must set that command unless there is no impact at that level.
-* Add the set module level alias(es) (`alias module=level"alias module module_level"`)
-* Possibly adjust presets in  `config/cfg/presets` to use the new module or levels to an existing module
+* Add the set module level alias(es) (`alias module=level"alias module module_level"`).
+* Possibly adjust presets in `config/cfg/presets` to use the new module or levels to an existing module.
 
 If you are adding a new module, you will also need to add a new `module` entry in `config/mastercomfig/cfg/comfig/modules_run.cfg`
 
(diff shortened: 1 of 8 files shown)


h9lk:
After update wanted so hard start translating fixes, perhaps need wait for your amendments and then i'll start)

mastercoms:
To clarify a few things, the spaces are added so that those values would appear within the autoexec template, which by default removes aliases. But in this case, I don't think the aliases are useful, since I remembered that the autoexec template is meant to be a supplement for mastercomfig customization.

The full autoexec is meant to be a transition from Comanglia's config and configs like it, and thus targets a similar feature level. This is why HUD achievements are disabled by default on it but not on any preset.

mastercoms (approved):

Tiagoquix:
@mastercoms Thank you very much for the clarifications!

Just a friendly reminder that you will have to delete the "mastercomfig no explosion smoke addon" manually on the dev release to make it go away. Not that this is related to this pull request, but I am taking the opportunity to say it.

Tiagoquix:
@mastercoms Another friendly reminder that you will have to delete the "user template" download if you want to only keep the "user autoexec template" rename (delete both on the dev and 9.7.0 releases).

mastercoms:
They're different things, no?

Tiagoquix:
> They're different things, no?

Oh, my bad. I'm really sorry for the confusion. I thought you had renamed "user template" to "user autoexec template", but no, [you renamed "autoexec template" to "user autoexec template".](https://github.com/mastercomfig/mastercomfig/compare/9.6.3...9.7.0#diff-423c3f1f691301df62711d00fb82ece59b74f2b80669219f62fb4e0666d7f573R43)

I ended up not checking it twice before making the comment. My memories confused me.

You don't need to do anything. Anyway, thanks for your attention and have a great day!

PR #619, opened 2022-03-15
Update quick_fixes.md

scrambl-d:
A user might have a config called `crosshairs_hitscan.cfg` that they would move to `cfg/user`. In their `/cfg/user/sniper.cfg` it would say `exec crosshairs_hitscan`, this would have to now read `exec user/crosshairs_hitscan`.


mastercoms:
I feel like it should be made more clear that its not required to move custom configs into user, it's just for if they happen to want to.

PR #622, opened 2022-03-17
fix: update ru localisation

h9lk:


h9lk:
@mastercoms Hello, noticed astrange duplication in source file, is this how it should be?
task

mastercoms:
No, line 88 shouldn't be there.

PR #626, opened 2022-03-21
Config updates

### LINKED ISSUE #625: Resolution of double running addons
I'm opening this issue to possibly find a solution to the addons system that went back to the old way, leaving aside the pre and post system.

The only problem I've seen so far is the fact that the Transparent Viewmodels addon executes the module commands after the modules are applied. This will, for example, cause a user who uses post_processing=high in their modules.cfg to be replaced with post_processing=off when using run_modules.

The simplest solution is to remove the second run of addons.

And, out of context, I noticed that you removed support for the move_bind module. I don't think it was on purpose, so I sent a pull request that changes the execution of the addons and adds the move_bind module again: #626

Tiagoquix:
This closes #625

### DIFF
--- a/config/cfg/addons/null-canceling-movement.cfg
+++ b/config/cfg/addons/null-canceling-movement.cfg
@@ -15,4 +15,20 @@ alias cb
 alias cl
 alias cr
 
+alias move_bind_wasd"bind w +mf;bind a +ml;bind s +mb;bind d +mr;alias move_bind_level echo move_bind=wasd"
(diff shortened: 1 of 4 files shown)


Tiagoquix:
@mastercoms Could you implement my changes made to modules.md?

Tiagoquix:
@mastercoms Could you implement my changes made to modules.md?

mastercoms:
no that module was removed, will put new instructions

PR #628, opened 2022-03-22
update ru localisation

h9lk:


treacherousfiend (changes requested):
The `user` folder is being replaced by a new `overrides` folder to reduce confusion with the `user.scr` file in `tf/cfg`

h9lk:
> The `user` folder is being replaced by a new `overrides` folder to reduce confusion with the `user.scr` file in `tf/cfg`

Was translating modules and didn't see that they changed another file that was translated before

treacherousfiend:
fair enough, should be a quick fix though

h9lk:
> The `user` folder is being replaced by a new `overrides` folder to reduce confusion with the `user.scr` file in `tf/cfg`


Was translating modules and didn't see that they changed another file that was translated before

> fair enough, should be a quick fix though

Don't have time to quickly fix this as doing translation myself (because only paths have been updated on your part - my translation has not been updated for 7 months and need to monitor changes in lines/parameters that have been added/changed during this time) and don't like small patches - think it's better to make big edits so that you can group everything together and put it in one commit

treacherousfiend (approved):

PR #648, opened 2022-06-26
improve native libraries section

getchoo:
currently the native libraries section seems a little lacking to me, especially in explaining why you would even want to use native libraries. there are good reasons, specifically fixing issues like [this](https://github.com/ValveSoftware/Source-1-Games/issues/1834#issuecomment-428303829) and i think they should be mentioned. 

i also added distribution specific launch options that forcefully pin native libraries via `LD_PRELOAD`. myself and others have ran into issues with TF2 falling back to the steam runtimes' libraries instead of the native versions mentioned here when only deleting the files from `bin`, and using these launch options seem to fix it.

### DIFF
--- a/docs/os/linux.md
+++ b/docs/os/linux.md
@@ -37,9 +37,22 @@ See the [Arch Linux wiki](https://wiki.archlinux.org/index.php/Sysctl#Virtual_me
 
 ## Native Libraries
 
-The automatic Steam runtime host library pinning is not enough to use native libraries
-on TF2 as the TF2 launcher script shades in some libraries.
+Using native libraries on Linux can have many performance benefits, as well as fix some issues
+such as mouse sensitivity not working as expected. The automatic Steam runtime host library pinning is
+not enough to use some native libraries on TF2 as the TF2 launcher script shades in some of its' own.
 
-Go to `TF2_FOLDER/bin` and delete `libSDL2-2.0.so.0`, and `libtcmalloc_minimal.so.4`.
-Now, Team Fortress 2 will not be able to load these libraries. The system will have to
-provide them, so make sure you have the 32-bit/multilib versions of `libtcmalloc` and `sdl2`.
+Two of the libraries that the game will benefit the most from using their native 
+version are `libSDL2` and `libtcmalloc`. To use your systems versions, go to `TF2_FOLDER/bin` and delete 
+`libSDL2-2.0.so.0`, and `libtcmalloc_minimal.so.4`. Now, Team Fortress 2 will not be able to load these 
+libraries and the system will have to provide them. Make sure you have `libtcmalloc` and `sdl2` installed, 
+as well as add one of the following lines to your launch options - depending on your distribution:
+
+Debian/Ubuntu-based distros:
+```
+LD_PRELOAD="/usr/lib/x86_64-linux-gnu/libSDL2-2.0.so.0:/usr/lib/x86_64-linux-gnu/libtcmalloc_minimal.so.4:$LD_PRELOAD" %command%
+```
+
+Arch/Fedora-based distros:
+```
+LD_PRELOAD="/usr/lib64/libSDL2-2.0.so.0:/usr/lib64/libtcmalloc_minimal.so:$LD_PRELOAD" %command%
+```
--- a/docs/os/linux.md
+++ b/docs/os/linux.md
@@ -48,11 +48,11 @@ libraries and the system will have to provide them. Make sure you have `libtcmal
 as well as add one of the following lines to your launch options - depending on your distribution:
 
 Debian/Ubuntu-based distros:
-```
+```sh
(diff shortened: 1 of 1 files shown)


mastercoms:
Revised it slightly, and did not include the `LD_PRELOAD` because bypassing Steam runtime libraries entirely is out of scope and can be done by using `STEAM_RUNTIME=0`. Doing it outside of that is not ideal. I also don't recommend disabling Steam runtime libraries or manually overriding Steam's `LD_PRELOAD`. You can use the pinning system instead. Also, just a heads up, even if this override was a good idea (and it's not), you should be including the previous value of `LD_PRELOAD` as well, or else it will break some stuff with Steam.

getchoo:
> did not include the `LD_PRELOAD` because bypassing Steam runtime libraries entirely is out of scope and can be done by using `STEAM_RUNTIME=0`. Doing it outside of that is not ideal.

this `LD_PRELOAD` launch option doesn't bypass all steam runtime libraries; it specifically preloads the two libraries mentioned in this part, and retains the previous `LD_PRELOAD` variable. now, it might not be ideal to use these two libraries specifically, but it does work and has been shown to resolve issues. it is is even recommend by the [arch wiki](https://wiki.archlinux.org/title/Steam/Game-specific_troubleshooting#Wrong_mouse_sensitivity) - which after considering how many other times it has been referenced in this doc, i think we could agree it's a good source.

>  I also don't recommend disabling Steam runtime libraries or manually overriding Steam's LD_PRELOAD. You can use the pinning system instead.

`LD_PRELOAD` is an extremely common way to fix issues on Steam for linux. it's all over protondb and (referencing it again :trollface:) the arch wiki mentions it [57 times](https://wiki.archlinux.org/title/Steam/Game-specific_troubleshooting) to fix issues in a lot different games. the `steam-native-runtime` package also does similar things even after using `STEAM_RUNTIME=0` by [editing](https://github.com/archlinux/svntogit-community/blob/packages/steam-native-runtime/trunk/steam-native.sh) `LD_LIBRARY_PATH` - which to me is arguably worse since replacing the entire runtime could cause more instability than overriding two specific libraries with great backwards compatibility. i think my solution might be a good middle ground between using most of the runtime and not using it at all.

> Also, just a heads up, even if this override was a good idea (and it's not), you should be including the previous value of LD_PRELOAD as well, or else it will break some stuff with Steam.

[i did](https://github.com/mastercomfig/mastercomfig/pull/648/commits/d39acb6ba6be8fb373d22f1c08ab5a8049b4f47e#diff-a0ef9b56121dd262a57dd5f11c6beb8ba1898420147128be151a8f789cbbd7c2R57)

overall, i can understand wanting to stick with the runtimes' libraries..but i really think it would destroy the purpose of a section called "native libraries" when the actual native libraries are not being used. a better option imo would be to change the sections' title to something like "force runtime libraries" as that would be more accurate under the current revision. alternatively, this pr could be continued if anything here convinced you that it would be acceptable to use native libraries. i think there would be good reasons to do this still, especially when considering fixes introduced in newer versions of libraries like SDL2 would most likely hit distros repositories first long before the steam runtime, and it would make for a better experience with those newer versions

mastercoms:
Did you miss the part where I said you can use the pinning system instead?

getchoo:
i don't see how pinning would be very different from what's going on here. besides that though, why not change the recommendation to something like
```sh
ln -s /usr/lib32/libSDL2-2.0.so.0 ~/.steam/root/ubuntu12_32/steam-runtime/pinned_libs_32
```

and still get the advantages of using actual native libraries?

mastercoms:
Because I haven't had the chance to test it

mastercoms:
If you can confirm it for me, you can PR it

getchoo:
i can try to test it myself and send a log or something if you’d like. the `LD_PRELOAD` launch option worked fine for me, so i have a feeling this would too - i’m just not sure if you would still need to delete the default libraries though (i know that part’s not a big deal, it just annoyed me :p)

mastercoms:
I'm pretty sure the ./hl2_linux.sh script which runs on launch will ignore it otherwise

getchoo:
just got it working on arch (using lib32-tcmalloc from aur)
```
$ sudo pldd (pgrep hl2_linux) | rg 'SDL|tcmalloc'
    /home/seth/.local/share/Steam/ubuntu12_32/steam-runtime/pinned_libs_32/libtcmalloc_minimal.so.4
    /home/seth/.local/share/Steam/ubuntu12_32/steam-runtime/pinned_libs_32/libSDL2-2.0.so.0
$ ls -l ~/.local/share/Steam/ubuntu12_32/steam-runtime/pinned_libs_32/ | rg 'SDL|tcmalloc'
    lrwxrwxrwx 27 seth 27 Jun 19:51 libSDL2-2.0.so.0 -> /usr/lib32/libSDL2-2.0.so.0
    lrwxrwxrwx 35 seth 27 Jun 20:29 libtcmalloc_minimal.so.4 -> /usr/lib32/libtcmalloc_minimal.so.4
```
making a new pr now

PR #649, opened 2022-06-28
improve native libraries section: electric boogaloo

getchoo:
continues #648 and a4ef83a48386f31020cb3c7ae846aae46abba3b9 by adding instructions to pin system libraries to the steam linux runtime

### DIFF
--- a/docs/os/linux.md
+++ b/docs/os/linux.md
@@ -40,24 +40,30 @@ See the [Arch Linux wiki](https://wiki.archlinux.org/index.php/Sysctl#Virtual_me
 
 ## Native Libraries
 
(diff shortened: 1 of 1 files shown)


mastercoms (dismissed):

getchoo:
- removed tcmalloc instructions
- added instructions for fedora
- added the name of the multilib packages for each distribution 
  - fedora's was hard to find, i imagine this would make it easier for a lot of people
- changed the wording of the main paragraph to explain how this works more
  - i think this guide should be more generalized; this way, if anyone wants to mess with other native libraries, they have enough information and a good example to follow in the SDL2 guide

mastercoms (approved):

PR #650, opened 2022-06-29
Some refinements, and some more information

felikcat:
docs/os/linux.md

### DIFF
--- a/docs/os/linux.md
+++ b/docs/os/linux.md
@@ -67,11 +67,3 @@ Fedora:
 sudo dnf install SDL2.i686
 ln -s /usr/lib/libSDL2-2.0.so.0 "$HOME/.steam/root/ubuntu12_32/steam-runtime/pinned_libs_32/"
 ```
-
-### Force SDL2 Wayland
-After using the native 32-bit SDL2 library, it's possible to run TF2 using its Wayland video driver. This could increase performance on GPUs where this works correctly (if any as of 29 Jun 2022), or rather when this is mature enough to use.
-
-Steam's "Library" -> Right-click "Team Fortress 2" -> "Properties..." -> Under "General" insert the following into "Launch Options": `SDL_VIDEODRIVER=wayland  %command% -your -tf2 -launch -options -go -here`
-
-NOTE: `SDL_VIDEODRIVER=wayland gamemoderun %command%` will also work.
-`%command%` specifies to set launch options for the Steam game; what comes before that is  Linux commands and environment variables.
--- a/docs/os/linux.md
+++ b/docs/os/linux.md
@@ -19,10 +19,12 @@ Use either the `performance`, `schedutil` or `ondemand` governors when playing T
 
 [GameMode](https://github.com/FeralInteractive/gamemode) is a handy utility released by Feral Interactive, which will change the CPU Governor from the default CPU governor to `performance`. Recent updates allow it to change a games' "nice priority", which will give the game more resources to work with.
 
-To run TF2 using GameMode through Steam, open Steam, then...
-Library -> right-click Team Fortress 2 -> Properties -> in the General tab, insert the following into your Launch Options: `gamemoderun %command% -your -tf2 -launch -options -go -here`
+To run TF2 using GameMode through Steam, open Steam, then:
+* Library -> right-click Team Fortress 2 -> Properties -> General -> Launch Options: \
+`gamemoderun %command% -your -tf2 -launch -options -go -here`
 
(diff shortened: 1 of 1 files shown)


docs/os/linux.md
```diff
@@ -67,3 +67,11 @@ Fedora:
 sudo dnf install SDL2.i686
 ln -s /usr/lib/libSDL2-2.0.so.0 "$HOME/.steam/root/ubuntu12_32/steam-runtime/pinned_libs_32/"
 ```
+
+### Force SDL2 Wayland
```
getchoo: as discussed in previous prs and commits, this doesn't actually work half of the time. testing under my own hardware - along with using many different versions of libSDL2 (steam runtime's, native, and the built-in lib) - this launches the game with a blank screen. this has also been reported [elsewhere](https://github.com/ValveSoftware/Source-1-Games/issues/3871)

even if this did run properly, a better recommendation here would be to use gamescope (which also doesn't work under the same conditions) as that is a more optimal way to run X11 games under xwayland. using gamescope is also more likely to work in the future over forcing `SDL_VIDEODRIVER` (work is being done [already](https://github.com/Plagman/gamescope/pull/454)), especially considering SDL2 on wayland is stable now..so it's not an issue on the library's side; it's a tf2 issue.

docs/os/linux.md
```diff
@@ -19,13 +19,17 @@ Use either the `performance`, `schedutil` or `ondemand` governors when playing T
...
+Steam's "Library" -> Right-click "Team Fortress 2" -> "Properties..." -> Under "General" insert the following into "Launch Options": `gamemoderun %command% -your -tf2 -launch -options -go -here`
+
 ## irqbalance
 
-Balance IRQ interrupts across multiple cores by installing the `irqbalance` package on your distro and enabling the `irqbalance` service. This package is installed on most debian-based OS's by default, like Ubuntu and Linux Mint.
+Balance IRQ interrupts across multiple cores by installing the `irqbalance` package on your distro and enabling the `irqbalance` service; likely: `sudo systemctl enable --now irqbalance.service`. 
+This package is installed on most debian-based OS's by default, like Ubuntu and Linux Mint.
 
-## Ananicy
+## Ananicy Cpp
```
getchoo, line 34: after looking at the gitlab page for ananicy-cpp, i am kinda interested - especially considering the original ananicy hasn't been getting commits in a while. if it still has the community rules support, this is awesome!

docs/os/linux.md
```diff
@@ -19,13 +19,17 @@ Use either the `performance`, `schedutil` or `ondemand` governors when playing T
 
 [GameMode](https://github.com/FeralInteractive/gamemode) is a handy utility released by Feral Interactive, which will change the CPU Governor from the default CPU governor to `performance`. Recent updates allow it to change a games' "nice priority", which will give the game more resources to work with.
 
+To run Team Fortress 2 using GameMode:
+Steam's "Library" -> Right-click "Team Fortress 2" -> "Properties..." -> Under "General" insert the following into "Launch Options": `gamemoderun %command% -your -tf2 -launch -options -go -here`
+
 ## irqbalance
 
-Balance IRQ interrupts across multiple cores by installing the `irqbalance` package on your distro and enabling the `irqbalance` service. This package is installed on most debian-based OS's by default, like Ubuntu and Linux Mint.
+Balance IRQ interrupts across multiple cores by installing the `irqbalance` package on your distro and enabling the `irqbalance` service; likely: `sudo systemctl enable --now irqbalance.service`.
```
getchoo: nitpick: this seems different from other styles used in places like line 32. maybe make the command more isolated (like in it's own line), or change it back to just mentioning the service name like other lines :)

docs/os/linux.md
felikcat: Doesn't launch with a blank screen for me on NVIDIA 515.48.07 with GNOME 42.2 Wayland. Graphically it's very broken, and the FPS is extremely low.

docs/os/linux.md
felikcat: https://gitlab.com/ananicy-cpp/ananicy-cpp#from-the-original-project

docs/os/linux.md
getchoo: that's pretty surprising considering i have the same setup, and you didn't mention using a native library or not so that would be pretty crazy if you weren't.

> Graphically it's very broken, and the FPS is extremely low.

and if this is the case, why add it? this will just break things for people. gamescope would still be a better alternative here but it has similar issues reported on nvidia

getchoo (dismissed):
i got no complaints, this is cool

mastercoms (approved):

PR #660, opened 2022-07-25
Fix spelling mistakes

micwoj92:


mastercoms:
Not accepting changes to cvarlist

micwoj92:
> Not accepting changes to cvarlist

Would you accept pr to team comtress 2 to fix these spelling mistakes there? I think this should count as user facing and therefore valid.

PR #667, opened 2022-08-21
Change anti-aliasing wording (release branch)

Tiagoquix:


mastercoms:
Fixed in 8ca5e16c63dd35bc408fdb0718df0efe4b994bcf

Tiagoquix:
> Fixed in [8ca5e16](https://github.com/mastercomfig/mastercomfig/commit/8ca5e16c63dd35bc408fdb0718df0efe4b994bcf)

Thanks! I suggest to change it to something like "multisample anti-aliasing (MSAA)" since "MSAA anti-aliasing" is the same as "multisample anti-aliasing anti-aliasing".

mastercoms:
It's a clarifying word, I think it's fine

PR #676, opened 2022-09-27
Addons Documentation

aSeriousDeveloper:
Added a documentation file for the current addons provided by Mastercomfig. Should mention most useful information to those installing Mastercomfig.


mastercoms:
Thanks, it's a nice idea, though I feel like this is a job for the app, presets are not documented either and selection is harder on the app if the information is not available there.

PR #690, opened 2022-11-22
Added a line to Killstreak Sheens noting that they do not work under DirectX 8

ghost:
Added a line stating that Killstreak Sheens don't work under DirectX 8, since I see alot of people using dx8 and wondering why their killstreak sheens aren't showing.


hyperventilation:
minor spelling mistake

micwoj92:
where

PR #695, opened 2022-12-22
Update linux.md

colias-palaeno:
- Added a warning for Steam Flatpak users (separate from the existing one for Wayland users) regarding using the system version of SDL2
- Fixed a dead link containing a Steam rules file for Anaincy

### DIFF
--- a/docs/os/linux.md
+++ b/docs/os/linux.md
@@ -58,6 +58,7 @@ Then, install a 32-bit version of SDL2 and pin the library to the runtime using 
 
 !!! warning
     When using Wayland, ```SDL_VIDEODRIVER=x11 %command%``` will need to be prepended to your launch options to avoid rendering issues
+    
 !!! warning
     Users of the Flatpak version of Steam will have to grant Steam access to the directory in which SDL2 is located. This can be done by setting the ```--filesystem``` Flatpak parameter which Steam uses to the relevant path.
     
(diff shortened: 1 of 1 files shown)


docs/os/linux.md
```diff
@@ -58,6 +58,9 @@ Then, install a 32-bit version of SDL2 and pin the library to the runtime using
 
 !!! warning
     When using Wayland, ```SDL_VIDEODRIVER=x11 %command%``` will need to be prepended to your launch options to avoid rendering issues
+!!! warning
```
mastercoms, line 62: Add another line between the 2 warnings

mastercoms (changes requested):

docs/os/linux.md
colias-palaeno: It's done. Should be fine to merge now

getchoo:
i don't think adding the warning for flatpak is the best way to go about this. while it is true that you need to give the flatpak permission to access the native library, it's a bit pointless since the flatpak version of steam already pins a (decently) up to date version of SDL from the runtime
```bash
$ realpath ~/.var/app/com.valvesoftware.Steam/data/Steam/ubuntu12_32/steam-runtime/pinned_libs_32/libSDL2-2.0.so.0
/home/seth/.var/app/com.valvesoftware.Steam/.local/share/Steam/ubuntu12_32/steam-runtime/usr/lib/i386-linux-gnu/libSDL2-2.0.so.0.2400.1
```

i think a better option here would be to make a note that only removing the built-in version SDL from `Team Fortress 2/bin` is required for the flatpak version

mastercoms (approved):

mastercoms:
@getchoo PR welcome

PR #696, opened 2022-12-24
add note for the steam flatpak in linux.md

getchoo:
continuation of #695

clarifies that the flatpak version of steam already pins a recent version of SDL2 to the runtime, so installing it with the system manager is unneeded


mastercoms (approved):

mastercoms:
good but wrong branch, should be release

getchoo:
🥸

mastercoms:
uhm, ACSHTUALLY

PR #698, opened 2023-01-17
Run "overrides/autoexec" after "app/autoexec"

Tiagoquix:
User manually created autoexec should override app's autoexec.


mastercoms:
App is something that has been created later by users

PR #699, opened 2023-01-17
Add missing semicolons

Tiagoquix:


mastercoms:
Don't need semicolons at the end of lines

Tiagoquix:
So the second semicolon [here](https://github.com/mastercomfig/mastercomfig/blob/develop/dev/presets/package.sh#L23) should be removed I think

mastercoms:
Sorry, misremembered how it's set up

mastercoms (approved):

PR #702, opened 2023-02-14
Clarify HUD avatars only works with floating health bars

Tiagoquix:


mastercoms (approved):

mastercoms (dismissed):

mastercoms:
They also show in the scoreboard.

Tiagoquix:
> They also show in the scoreboard.

This is irrelevant to this module, because HUD avatars only affect **Target ID** (i.e. mouse over a player to see their nickname and photo). On the scoreboard, the avatars are always shown (even if HUD avatars are disabled).

Floating health bars are enabled by default, but can be disabled, and when disabled, they hide the avatars.

mastercoms:
Yes, but when you say it like in the PR, I think confusion comes up, what it says right now with just saying it's target ID avatars, it's clear enough.

Tiagoquix:
Sure, I'll do another PR to make it more clear.

mastercoms:
It's already clear: it says it controls the display of avatars for target ID. When you add an additional condition for when it "controls" that, then people can get confused.

Tiagoquix:
Yes, but the target ID avatars aren't displayed when floating health bars are enabled.

Tiagoquix:
I see. We could rephrase it as:

"Controls the display of target ID avatars. Target ID only works when floating health bars are enabled."

PR #704, opened 2023-03-05
improve wording about anti-aliasing on Linux

Ristovski:
Previously it read as if the AMD drivers were to blame for the slow full clear, where in reality it is TF2 causing it


mastercoms:
Pushed in f270c7255b943a05fc644df8c56c174659f2c505

PR #726, opened 2023-09-05
docs(customization): add spaces to alias examples

rufio-tf2:
Some of the examples for redefining aliases in the Customization documentation for Custom Configs don't have a space between the alias name and the quoted command string.

I'm assuming that not having a space there was incidental, rather than intentional.

These changes are trivial, but the spaces help me read it more easily, and perhaps will help others too.

Example:
```diff
-alias class_config_heavyweapons"exec overrides/myheavy.cfg"
+alias class_config_heavyweapons "exec overrides/myheavy.cfg"
```


mastercoms:
It's intentional but if there are readability concerns then this can be considered

rufio-tf2:
> It's intentional but if there are readability concerns then this can be considered

Sounds good. I'll let you close it whenever you want.

mastercoms (approved):

PR #737, opened 2024-02-11
Add warning against simultaneous use of Ananicy Cpp and Gamemode

bigrand:
These two programs basically do the same thing. Using them simultaneously is counterproductive. You must choose between allowing Gamemode to adjust the nice level for TF2 or letting Ananicy Cpp do it instead. Mixing the two may lead to instability and suboptimal performance.

### DIFF
--- a/docs/os/linux.md
+++ b/docs/os/linux.md
@@ -35,6 +35,9 @@ This package is installed on most Debian-based OS by default, like Ubuntu and Li
 
 ## Ananicy Cpp
 
+!!! warning
+    Ananicy Cpp and Gamemode each modify the nice level of processes. Using both simultaneously could lead to potential instability issues.
+
 You can install `schedtool` and [Ananicy Cpp](https://gitlab.com/ananicy-cpp/ananicy-cpp), then enable `ananicy-cpp.service` to automatically apply rule-based process priority balancing, improving resources allocated to TF2.
 
 This, however, requires a [profile for TF2](https://github.com/Nefelim4ag/Ananicy/blob/master/ananicy.d/00-default/games/_steam.rules). Follow Ananicy Cpp's [profile import instructions](https://gitlab.com/ananicy-cpp/ananicy-cpp#community-rules) to use community profiles from the original Ananicy software, which can be downloaded [here](https://github.com/Nefelim4ag/Ananicy) or by installing the original Ananicy.


getchoo:
why not remove ananicy? it's a bit weird to have two pieces of software on this list that do the same thing, and gamemode is much easier to use with it having no profile requirement and already being commonly used by other games and apps - it's also more maintained in general

mastercoms:
isn't ananicy affecting much more than gamemode?

mastercoms:
this is the reason why we use both, because ananicy can affect other priorities and other settings, and gamemode can cover for things that ananicy doesn't cover

bigrand:
> this is the reason why we use both, because ananicy can affect other priorities and other settings, and gamemode can cover for things that ananicy doesn't cover

The features that really matter overlap because they are managed by both programs. You can use Ananicy CPP and Gamemode at the same time, but you need to configure them so that one of them takes care of I/O priority, nice level, and CPU scheduling. For all that matters, Gamemode does it more intuitively.

Ananicy CPP automatically readjusts the nice level and CPU priority of processes. Its main job is similar to what Gamemode already does, but instead of focusing only on games, it applies to other apps as well. To use Ananicy CPP, you need to write rules or install community ones whereas Gamemode runs out of the box.

If you read the Ananicy CPP community-maintained rules (which are mostly maintained by the CachyOS team, btw they also advise to not use Gamemode and Ananicy CPP at the same time), you’ll notice that they modify the following parameters for games:
`{ "type": "Game", "nice": -7, "ioclass": "best-effort", "latency_nice": -7 }`

Quick breakdown:
-`nicelevel`: Gamemode already does this.
-`ioclass`: According to the Ananicy CPP repository, this is only supported by the legacy I/O scheduler CFQ, which was removed from the Linux kernel back in version 5.0.
-`latency_nice`: This feature requires a patch that hasn’t been merged yet (I think), so you’d need a custom kernel to use it.

Ananicy CPP has a few more attributes to set up but I don't think they are relevant to performance. So, unless you specifically require features that the majority of people won't have installed, enabled, or even be interested in using, Ananicy CPP would essentially be a harder to set up Gamemode. For example, I wouldn't use `ioclass` because BFQ I/O scheduler works way better for my use case than CFQ.

I really don't care if Ananicy CPP stays in the docs or not, but if it does, there definitely should be a warning.

Links:
[Ananicy CPP Community Maintained Rules](https://github.com/CachyOS/ananicy-rules)
[CFQ Removal in Linux Kernel](https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=f382fb0bcef4c37dc049e9f6963e3baf204d815c)
[latency_nice patch](https://lore.kernel.org/lkml/20220322163911.3jge4unswuap3pjh@wubuntu/T/)

Archived related discussion which may explain it better than I do:
https://github.com/Nefelim4ag/Ananicy/issues/120

mastercoms:
so controlling the nice levels of other applications doesn't help with game performance?

bigrand:
I suppose so. Although I don't believe it would make much of a difference because programs that require a large amount of resources will continue to require a large amount, and those that don't would most likely make little difference, but I don't have any benchmarks to back it up. I've always used Ananicy CPP on PCs that can easily slow down because, in my experience, it works better to improve overall system responsiveness than raw process performance.

I still believe it's important to include a warning to inform users that using the same features that both programs share for TF2's process may cause instability.

btw, if Ananicy CPP remains in the docs (most probable), the Ananicy CPP community rules link should be updated to go to the CachyOS team's ones, as the old repo has been archived.

mastercoms:
what instability does it cause? do they constantly set the nice level and fight over it?

bigrand:
According to the Founder of CachyOS (I asked in their Discord), it appears to be undefined behavior, which means it could be anything from nothing happening to random crashes. He did say that one will overwrite the other's nice value, but it's pointless to use both at the same time. I tried to ask him some more questions, but he didn't respond.

Based on what he said, I believe there isn't much of a difference between the two, so choosing one over the other is a matter of preference. You can use Ananicy CPP and Gamemode at the same time, but you need to remove the rules or maybe use only the settings that are exclusive to each program, but I'm not sure, it seems too complicated for probably no benefits at all IMO.

Maybe the warning should be rewritten to mention the undefined behavior?

mastercoms:
Undefined behavior in this case meaning either one nice level gets set or the other? As long as they're not setting them continuously I can't see how it would cause instability, but I can see how it would be pointless to recommend installing Steam rules for Ananicy.

bigrand:
I'm not sure. I created this issue because of the several warnings I found while searching about several issues I was having with Ananicy CPP (it was interfering with my login session), so I'm not particularly familiar with how Ananicy CPP or Gamemode work internally.

For what it's worth, here's what the CachyOS founder said when I asked about it:
>what happens when you use Ananicy CPP and Gamemode at the same time? ... won't one override the other without issue?
"one will overwrite the other, and vice versa every time [a] new task is created"

>So there are no stability issues if you use both?
"it's UB
anyways using both at the same time doesnt make any sense
also gamemode for most games just sets sched to performance mode"

PR #747, opened 2024-05-08
Add info about setting up drirc file to avoid black artifacts when using mat_phong 0 and such

rrkkss:
added for both english and russian


Tiagoquix:
For doc changes, target the `release` branch instead.

May be worth noting that this doesn't seem to happen with DXVK (it seems to only happen with ToGL), but this is not confirmed or denied.

rrkkss:
well, I based my branch off development, so yeah there'll be some rebasing

anyways, I got it to run with dxvk, that's why I made the PR in the first place

mastercoms:
This isn't much of a priority anymore due to Vulkan.

mastercoms:
Manually merged in https://github.com/mastercomfig/mastercomfig/compare/900f657f5aad...2c9dab7f2037

PR #757, opened 2024-06-26
Config updates

### LINKED ISSUE #755: Restore default values for some HUD-related ConVars
Hi there.

There are some ConVars that have different values compared to TF2 default in mastercomfig, most of them being related to the heads-up display - such as:

- hud_saytext_time
- cl_notifications_move_time
- hud_deathnotice_time

etc.

In the past, I saw then as somewhat interesting to reduce HUD elements and therefore improve FPS somewhat and make the screen cleaner with less elements taking space for too much time. However, now, some years leater, I don't see a reason for us to set different values. I think we should use the default values whenever possible if it doesn't hurt user experience (for example, tf_dashboard_slide_time makes sense to be a lower value due to it improving user experience with fast transitions).

Please let me know what you think. If the suggestion is approved, I'll do a PR myself to change the ConVars' values.

Thanks in advance for your attention.

### LINKED ISSUE #756: Question about network ConVars
`rate`: considering TF2 updates, can we increase the default `bandwidth` module level to `1.5Mbps`?

`cl_pred_optimize`: should we use `setinfo` for it like we use for `cl_interp` and `cl_interp_ratio`?

Tiagoquix:
- change `sheens_speed` default to `medium`
- add `hud_combattext 0` to `destitute` preset
- add `setinfo` for `cl_pred_optimize`
- change some comments
- uncomment `hud_combattext_healing 1` and enforce it
- set default value for: `hud_saytext_time`, `hud_deathnotice_time` and `cl_hud_killstreak_display_time`
- comment `cc_captiontrace 0`; the config, IMO, should not touch CC-related captions, they should manually be changed by the user (such as `cc_predisplay_time` from `0.25` to `0`)
- add another replay cvar

please let me know if I missed something. thanks!

Closes #755
Closes #756


Tiagoquix:
if it is preferred for me to base on staging instead of develop, please let me know. thanks!

mastercoms:
* why does the default for sheens speed need to change?
* why do we need to enforce healing text on?
* that cvar adjusts output for missing closed captions so they don't show up in the console if they're missing
* that replay cvar wasn't added because its a render option in the UI. the others control hidden options which are persistent preferences.

Tiagoquix:
> why does the default for sheens speed need to change?

while I was resetting the HUD ConVars default values, I thought that the sheens speed could be reverted too, as I consider that the current one defined by mastercomfig to be more of a "good value" rather than a "definite improvement", so to speak

there's not really a reason, but please let me know what you prefer and I'll be fine either way

> why do we need to enforce healing text on?

in the same way that the Crit text is forced to display, I thought it would also be useful to force the healing text to be displayed

> that cvar adjusts output for missing closed captions so they don't show up in the console if they're missing

yes, this is useful for users who use custom captions (such as, for example, some captions not covered by clovervidia captions), but this can hinder subtitle developers as errors do not appear in the console

thinking about it now, I agree that this is very niche and I agree with keeping the value at 0

---------------------

I have made some changes. please let me know what you think. thanks!

PR #762, opened 2024-07-20
Add User-Generated Decals module

### LINKED ISSUE #759: Incorporate custom decals into sprays module
With the introduction of the ConVar `tf_hide_custom_decals`, I think that the `sprays` module could be renamed to `custom_decals` (or `user_decals`) and then incorporate both sprays and user decals (Cons. Objector, flair etc.) into a single module.

Tiagoquix:
This closes #759.

I'm not sure if the `cl_allowupload` and `cl_allowdownload` commands don't actually affect custom decal images (such as Conscientious Objector etc.), but I assume they don't.

This PR follows TF2 defaults. Default level for this new module is the same as TF2 default values.


Tiagoquix:
Updated:

- Now Very Low and Low presets use `decals_user=off` instead of `decals_user=low`.
- Correct module levels in `modules.json`.

Tiagoquix:
As much as I see it as possible for new module levels in which temporary files are kept, and either only sprays or only custom decals are allowed, I don't think it would make much sense, so I don't plan to include these new levels. In my view, if you are going "as far" as keeping temporary files, then it most likely means you want to have everything enabled.

Even though we don't provide it, users can still use their own overrides in their `autoexec.cfg`.

PR #777, opened 2025-02-20
Config updates

Tiagoquix:
Various random changes throughout many config. files.

# General
2. Fixed inconsistencies between `module_levels.cfg` and `modules_run.cfg`.
4. Miscellaneous comments changes/additions/improvements for config and docs.
5. Copied changes from `comfig.cfg` to `config_template.cfg`.
6. `ragdols=hidden` and `ragdolls=low` are now aliases for `ragdolls=off`.
7. Added `decals_models=medium` (max. model decals 9). Documentation for low level updated to reflect latest changes.
8. Make mastercomfig not set `tf_quest_map_tuner_wobble_magnitude` anymore. Only Very Low sets this, but it doesn't have `FCVAR_ARCHIVE`, and the default is the same value set by mastercomfig, so no need to set it.
9. Fixed `texture_filter_blocky` not being correctly aliased.
10. Added missing commands to `texture_filter_trilinear`.
13. Fixed invalid `very_low` module level for the `texture_quality` module.
14. Fixed incorrect default value for `bandwidth` in docs after recent change.
17. Added recommendation for `snapshot_buffer=auto` in docs and comfig app.
18. Renamed some leftovers of `post_comfig.cfg` to `setup_hook.cfg`.
19. Misc. adjustments to Lighting Ex documentation.

# Presets
21. All presets are using `texture_quality=high` and I don't know why. This doesn't make much sense for me, so I have adjusted the presets values accordingly. I have also adjusted `texture_filter` for some of them.
22. Very low preset only: sorted all extra commands alphabetically. Added `hud_combattext 0` and `viewmodel_fov_demo 0`.
23. Very low reset CFG: add previous commands.

# OpenGL
24. Consolidated OpenGL optimizations into a single file. Adjusted documentation regarding ToGL.
25. OpenGL `lighting_ex=high` setting could be overriden by presets. I've added a new file `legacy_tweaks.cfg` that ensures OpenGL optimizations have priority over presets, but still respect user-defined overrides.

# Matchmaking Override (MM Override) CFG
26. Added, adjusted and removed some commands.
- Not set by mastercomfig anymore (therefore, removed):
  - `filesystem_buffer_size`, `r_dopixelvisibility` and `r_pixelfog`.
- Adjusted:
  - `fps_max` default is now `400`.
  - Ensure some commands are 100% the default value by adding zeros.
- Removed:
  - `mat_trilinear` (won't cause crashes in my opinion, so not that important).
  - `net_splitpacket_maxrate` (new comment in `comfig.cfg` from `staging` branch says it should be the same value as the `rate` ConVar, and `rate` is not reset in `mm_override.cfg`).
- Added:
  - `filesystem_max_stdio_read` and set it to the default value (mastercomfig sets it to a custom value).
  - `anim_3wayblend` and `lod_TransitionDist`; I believe they are important, so I have added them.
27. Added new echo to clarify that game settings persist after the match.

CC @mastercoms - please review before the next stable release. Many thanks! :)


Tiagoquix:
Added `sv_use_steam_networking 0` to `local_server=deny` (just to be extra safe) after I saw https://www.reddit.com/r/tf2/comments/1its8wr/so_i_have_been_a_victim_of_my_server_being_public/.

mastercoms:
Bilinear was removed because it was useless, there's really no one who benefits performance-wise from this setting and people ask all the time about the artifacting it produces, to then be told to switch off of bilinear. Blocky filtering is an actually desired look + performance level.

Texture quality is kept at high to go back to an old mastercomfig optimization (mat_picmip 0 has highest performance) instead of putting low quality textures for the sake of it.

Steam Networking shouldn't be a setting. It helps hide your IP, like the previous -enablefakeip option that was recommended. I am aware that people can join your game so I'm working on a password generation system instead, but due to not listing the server, it isn't as bad.

No, we aren't going to immediately break a module level before releasing it for the first time. Ragdolls should be kept.

Very Low reset comes with the assumption you're running it when you have already switched away from Very Low, so the new message isn't needed. It doesn't make sense to use it otherwise.

Why sort commands alphabetically? There was a logical order to them before, it doesn't make sense to me why it'd be changed.

The rest of the changes I'll review later

Tiagoquix:
In regards to...

**Bilinear Filtering**, it is a setting in the advanced graphics options, so that's why I suggest to keep it. Though, I agree with your arguments, so I'm not sure if there's too much benefit to keep it.

**Blocky filtering**, I have no intentions to remove it.

**Texture quality levels**, maybe we should use the `high` level in everything except the Ultra preset for even more optimization (i.e. also affect Medium-High and High presets).

**Steam Networking**, yes, it helps hide your IP, but as I understand it, disabling it while also using `sv_lan 1` etc. makes it so one of the possible ways to connect to your server is disabled. I think it's better, for overall security, not to allow Steam Networking to function, so we use an extra command to protect users even more. (Related upstream issue: https://github.com/ValveSoftware/Source-1-Games/issues/7092)

**`ragdolls=low`**, that's fine, this change will be reverted. We should mention in the release notes that is being deprecated.

**Very Low Reset CFG message**, in theory it is possible to use that and then restart your game to apply the new, post-removal settings, so that's why I added that message. Please let me know if this makes sense for you, and if it doesn't, then I'll remove that message.

**Alphabetically sorting**, it's because I think it makes it better to read, but your argument is fair, so I'll revert those changes while keeping the new commands.


I'll wait for your full review before doing any new commit to make new changes. Thanks.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -119,12 +119,12 @@ alias packet_size packet_size_large
 // Rate is recommended to be 80% of the server route speed
 // Download speed is preferred, since client-to-server networking is minimal
 
-net_maxcleartime 0.01 // cmdrate is the bound, keep this at minimum
+net_maxcleartime 0.01 // Cmdrate is the bound, keep this at minimum
```
mastercoms: cmdrate cannot be capitalized

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -240,8 +238,11 @@ alias lod
 //mat_reducefillrate 0 // Increase shading quality
 //mat_reducefillrate 1 // Simplify material shading and use some DirectX 8 features
 //r_dopixelvisibility 1 // Use pixel visibility system instead of line traces
-//r_pixelvisibility_partial 1 // Use partial visibility algorithm
+//r_pixelvisibility_partial 1 // Use a partial visibility algorithm
+//r_pixelvisibility_partial 0 // Do not use a partial visibility algorithm
 //building_cubemaps 1 // Force some lighting effects off
+//mat_phong 1 // Use a phong shader for shading/reflection
```
mastercoms: mat_phong is intentionally next to rimlight

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -240,8 +238,11 @@ alias lod
 //mat_reducefillrate 0 // Increase shading quality
 //mat_reducefillrate 1 // Simplify material shading and use some DirectX 8 features
 //r_dopixelvisibility 1 // Use pixel visibility system instead of line traces
-//r_pixelvisibility_partial 1 // Use partial visibility algorithm
+//r_pixelvisibility_partial 1 // Use a partial visibility algorithm
+//r_pixelvisibility_partial 0 // Do not use a partial visibility algorithm
```
mastercoms: No need to do this kind of duplicate stuff for clear toggles that don't have extra information besides the inverse. For example look above for examples of good toggle descriptions which are needed. It's not needed for this one.

image

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -264,9 +265,9 @@ alias lighting
 // Changing these variables from default cause full material reloads
 // which may be harmful to performance and also cause bugs
 
-alias lighting_ex_low"mat_bumpmap 0;mat_specular 0;mat_phong 0;alias lighting_ex_level echo lighting_ex=low"
-alias lighting_ex_medium"mat_bumpmap 1;mat_specular 1;mat_phong 0;alias lighting_ex_level echo lighting_ex=medium"
-alias lighting_ex_high"mat_bumpmap 1;mat_specular 1;mat_phong 1;alias lighting_ex_level echo lighting_ex=high"
+alias lighting_ex_low"mat_phong 0;mat_bumpmap 0;mat_specular 0;alias lighting_ex_level echo lighting_ex=low"
```
mastercoms: The order is intentional

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -597,10 +598,10 @@ alias decals
 
 //r_drawmodeldecals 1 // Enable blood decals for better determining how hurt someone is
 //r_drawmodeldecals 0 // Disable blood decals on bodies. Will prevent new objects to be created that decals apply onto. This also causes a bug with red bullet decals on some wood models.
-//r_maxmodeldecal 1 // Optimize maximum model decal count
+//r_maxmodeldecal 9 // Optimize maximum model decal count
 
 alias decals_models_off"r_drawmodeldecals 0;alias decals_models_level echo decals_models=off"
-alias decals_models_low"r_drawmodeldecals 1;r_maxmodeldecal 1;alias decals_models_level echo decals_models=low"
+alias decals_models_low"r_drawmodeldecals 1;r_maxmodeldecal 9;alias decals_models_level echo decals_models=low"
```
mastercoms: Why is this 9?

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Added that because in `config_template.cfg` it is present without any comments, and looks a bit off

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: > `r_maxmodeldecal` has been restored to `9` for `decals_models=low`. `1` would basically be the equivalent of having the module disabled, and `9` is the value used for `decals_low`. (If this change is ignored, the docs need to be updated for the new value of `1`.)

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1054,12 +1059,12 @@ alias match_hud match_hud_on
...
 
 alias dynamic_background_off"alias dynamic_background_level echo dynamic_background=off"
-alias dynamic_background_preload"map_background preload_room;wait 5;checkpointservercommand;script_execute randommenumusic;wait 5;disconnect;playmenumusic;alias dynamic_background_level echo dynamic_background=preload"
-alias dynamic_background_itemtest"map_background itemtest;wait 5;checkpointservercommand;script_execute randommenumusic;wait 5; disconnect;playmenumusic;alias dynamic_background_level echo dynamic_background=itemtest"
-alias dynamic_background_dustbowl"map_background background01;wait 5;checkpointservercommand;script_execute randommenumusic;wait 995;stop;playmenumusic;alias dynamic_background_level echo dynamic_background=dustbowl"
+alias dynamic_background_preload"sv_allow_wait_command 1;map_background preload_room;wait 5;checkpointservercommand;script_execute randommenumusic;wait 5;disconnect;playmenumusic;alias dynamic_background_level echo dynamic_background=preload"
```
mastercoms: This is already default 1

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1196,8 +1201,8 @@ alias hud_contracts hud_contracts_custom
 // --------------------
 // Achievement effects and HUD tracker
 
-//hud_achievement_count 8 // Max number of achievements that can be shown on the HUD
-//hud_achievement_count_engineer 3 // Max number of achievements that can be shown on the HUD when playing as Engineer
+//hud_achievement_count 8 // Max. number of achievements that can be shown on the HUD
```
mastercoms: I don't like these periods here

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: Yes, but the user may have a script or config that sets it to `0` somehow, so that's why I added to guarantee it will work. However, I kind of agree that this is exaggeration.

config/mastercomfig/cfg/comfig/comfig.cfg
```diff
@@ -1915,7 +1920,7 @@ alias apply_overrides"modules_c;run_modules;exec overrides/autoexec.cfg;exec app
 alias restore_preset"exec comfig/comfig.cfg;preset;run_modules"
 alias restore_config"exec autoexec.cfg"
 
-alias run_game_overrides"exec comfig/game_overrides.cfg;lightmaps_override"
```
mastercoms: This is meant to run twice.

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: It's to abbreviate "Maximum", but fine, I'll remove them

config/mastercomfig/cfg/comfig/legacy/opengl.cfg
```diff
@@ -1,8 +1,11 @@
-mat_bufferprimitives 0 // OpenGL must flush out primitives
-gl_pow2_tempmem 1 // Use better texture storage for better memory usage/fragmentation
-mat_forcehardwaresync 0 // ^
-gl_debug_output 0 // Disable debug output
-gl_magnify_resolve_mode 1 // Fastest upscale mode, instead of nicest
 gl_batch_tex_creates 1 // Use a reserved list of gl textures to batch
```
mastercoms: Please don't reorder this file

dev/presets/package.sh
```diff
@@ -20,9 +20,10 @@ for F in ../../config/cfg/presets/*; do
         printf "preset=%s;" "${P}"
         printf "exec overrides/pre_comfig.cfg;"
         printf "exec comfig/comfig.cfg;"
-        printf "exec overrides/setup_hook.cfg;"
+        printf "exec overrides/post_comfig.cfg;"
```
mastercoms: Why was this renamed?

docs/customization/modules.md
```diff
@@ -191,14 +191,17 @@ Default setting: based on which preset you are currently using.
...
 - **CPU usage:** low
 - **GPU usage:** high
 
 Default setting: based on which preset you are currently using.
 
-- **`lighting_ex=low`**: Better hints to use the lighting fast path (disables bumpmaps, specular and phong). May reduce performance on modern PCs, due to rendering reloads.
-- **`lighting_ex=medium`**: Disables phong, but keeps material paths for specular and bumpmaps, as disabling these can cause glitches on DX9. May reduce performance on modern PCs, due to rendering reloads.
-- **`lighting_ex=high`**: Uses whatever the material's lighting needs (enables bumpmaps, specular and phong).
+- **`lighting_ex=low`**: Better hints to use the lighting fast path (disables phong, bumpmaps and specular). May reduce performance on modern PCs due to rendering reloads.
+- **`lighting_ex=medium`**: Disables phong, but keeps material paths for bumpmaps and specular, as disabling these can cause glitches on Direct3D 9. May reduce performance on modern PCs due to rendering reloads.
```
mastercoms: DX9 is correct

docs/customization/modules.md
```diff
@@ -191,14 +191,17 @@ Default setting: based on which preset you are currently using.
...
+
 - **CPU usage:** low
 - **GPU usage:** high
 
 Default setting: based on which preset you are currently using.
 
-- **`lighting_ex=low`**: Better hints to use the lighting fast path (disables bumpmaps, specular and phong). May reduce performance on modern PCs, due to rendering reloads.
-- **`lighting_ex=medium`**: Disables phong, but keeps material paths for specular and bumpmaps, as disabling these can cause glitches on DX9. May reduce performance on modern PCs, due to rendering reloads.
-- **`lighting_ex=high`**: Uses whatever the material's lighting needs (enables bumpmaps, specular and phong).
+- **`lighting_ex=low`**: Better hints to use the lighting fast path (disables phong, bumpmaps and specular). May reduce performance on modern PCs due to rendering reloads.
```
mastercoms: why were these re-ordered?

docs/customization/modules.md
```diff
@@ -888,7 +891,7 @@ Default setting: **`mod_support=on`** (all presets).
 Controls the privacy of your Casual/Competitive matchmaking party.
 
 !!! warning
-    The Friends Only restriction for parties can be bypassed, which means anyone can join your party if they have your Steam ID. Leaving your party open can result in anyone to join your party and spam party chat, causing lag, sound spam and other disruptions.
+    The Friends-Only restriction for parties can be bypassed, which means anyone can join your party if they have your Steam ID. Leaving your party open can result in anyone to join your party and spam party chat, causing lag, sound spam and other disruptions.
```
mastercoms, line 866: Does it have a hyphen in game?

dev/presets/package.sh
Tiagoquix: Because it currently is renamed. See https://github.com/mastercomfig/mastercomfig/tree/staging/config/templates/overrides

I just finished the changes. Also, new documentation regarding OpenGL optimizations mentions post_comfig.cfg and not setup_hook.cfg

mastercoms:
> Texture quality levels, maybe we should use the high level in everything except the Ultra preset for even more optimization (i.e. also affect Medium-High and High presets).

At those levels, the quality/performance tradeoff that the user desires makes it so keeping it at only mat_picmip 0 isn't worth it.

> Steam Networking

Yes but all the settings in tandem prevent the common case of joining someone's game from the server browser at least. The rest will be done through password setting so you can't even join by ID somehow. Hiding the IP is very important to privacy.

> Very Low Reset CFG message, in theory it is possible to use that and then restart your game to apply the new, post-removal settings, so that's why I added that message. Please let me know if this makes sense for you, and if it doesn't, then I'll remove that message.

I feel like it just adds confusion. Usually people try very low preset, get disgusted, exit the game, switch preset, and now need a way to reset the lingering settings.

docs/customization/modules.md
Tiagoquix: DirectX 9 is the conjuction of APIs that make applications and games work, while Direct3D is the one responsible for rendering. I can't argue against you here, so I agree. That change will be reverted.

docs/customization/modules.md
Tiagoquix: Personal preference. I always read/feel like that when people mention those three commands they mention is this order: phong, bumpmap and specular. Current order also seems to be random (not alphabetically ordered, for example). Is there any reason to prefer the current order?

docs/customization/modules.md
Niterux: It doesn't appear to be referred to as "Friends-Only" or "Friends Only" in-game at all
"tf_party_join_request_mode" = "0" min. -1.000000 ( def. "-1" ) max. 2.000000 archive

 client - The current mode for incoming party join requests:
  -1 - Use default (currently 0), 
   0 - Open to friends, 
   1 - Friends can request to join, 
   2 - Invite only
image

Tiagoquix: Doesn't seem like the game uses either "Friends-Only" or "Friends Only".

image
image
image

config/mastercomfig/cfg/comfig/comfig.cfg
Tiagoquix: @mastercoms same discussion as https://github.com/mastercomfig/mastercomfig/pull/777#discussion_r1972497645

Tiagoquix:
> Hiding the IP is very important to privacy.

Yes, but if you can't connect to Steam Networking, then your IP won't even be announced to Steam servers somehow. If the user wants to allow remote connections with `local_server=allow`, we will enforce Steam Networking to ON.

I don't think the password generation system is necessary. It's cool, yeah, but I don't think it's worth it. The commands that we are using in `local_server=deny` are enough for this.

The idea to disabling Steam Networking too is to provide an extra layer of protection against remote connections by default, which is the purpose of the module.

Could you clarify what you mean by "joining by ID"? I think that even with the current console commands (without including Steam Networking) this is already not possible for local servers (unless you're talking about party invites).

------------

To the rest I didn't quote from the last comment: I agree and will be changing those.

docs/customization/modules.md
Tiagoquix: @Niterux I commented before refreshing the page, so I didn't see your comment xD
Nice catch too

mastercoms:
> Yes, but if you can't connect to Steam Networking, then your IP won't even be announced to Steam servers somehow. 

That is false. Every server is registered with Steam game servers, your IP will be visible in game info. It's not extra protection to disable Steam Networking, it's less protection. The password system is necessary for the maximum protection.

Tiagoquix:
@mastercoms Could you take a look at my responses in some of your reviews so I can take a look to commit new changes? Thanks.

docs/customization/modules.md
mastercoms: So just have it be Friends, not Friends Only

docs/customization/modules.md
mastercoms: It's just something english does to prefer certain words in a list based on their structure, syllables, sounds, etc.

dev/presets/package.sh
mastercoms: image

config/mastercomfig/cfg/comfig/comfig.cfg
mastercoms: Yes it is, we never set defaults for non-archive cvars

config/mastercomfig/cfg/comfig/comfig.cfg
mastercoms: It's a different discussion, the order here is due to how the shader combos work, I prefer to disable them in this order

config/mastercomfig/cfg/comfig/comfig.cfg
mastercoms: Yes, update the docs. There isn't a reason to have more than 1, for performance reasons. If you want to add a medium setting with 9, go ahead though

config/mastercomfig/cfg/comfig/comfig.cfg
mastercoms: That's okay, config_template.cfg is meant to be a file for people who use text cfgs, which has stuff like that for quick uncommenting. Please don't do this, there's no need for a comment

dev/presets/package.sh
Tiagoquix: That's strange, but OK. I'll rename any remaining instance of `post_comfig.cfg` to `setup_hook.cfg`.

docs/customization/modules.md
Tiagoquix: I think it sounds weird to say "The Friends restriction", so I'll keep the "only" or reword it.

"The Friends Only restriction" without a hyphen sounds strange because `Only` can be interpreted as part of `restriction`.

Tiagoquix:
@mastercoms 

Some questions I have before making new changes:

1. Regarding Bilinear Filtering, can we keep it for the Very Low setting since it is supposed to be the lowest possible and squeeze out every tiny bit of performance for the user? I would add a note to the docs to inform users that it can cause rendering bugs, so it would make users avoid it.
2. Regarding the Very Low preset, can we keep the lowest texture quality for it? As I understand it, for the Very Low preset at least, it makes more sense to keep the lowest texture quality to reduce VRAM usage, even though it may not be the optimal setting in regards to performance improvements. I believe it makes more sense for the Very Low preset to have the lowest texture quality.
3. Regarding the new changes I'll make: do you prefer that I make it in separate commits and/or PRs, or is it fine to just make a one big commit and then you take a look at "Files changed" afterwards?

mastercoms:
1. I don't really think bilinear filtering actually improves performance on 99% of PC hardware, especially in TF2
2. I don't think we should make a decision about VRAM about the cost of performance for very low, that isn't what very low is.
3. It doesn't matter, you can do either, I look at each commit if available and overall.

Tiagoquix:
Thanks. I will make the adjustements in the near future in a single commit, since the overall "Files changed" will be much less then.

docs/customization/modules.md
Tiagoquix: I will keep the hyphen.

Tiagoquix:
@mastercoms Hi! Sorry for the delay.

I have finalized all the requested changes and also done some more:

- Transferred some changes from `comfig.cfg` to `config_template.cfg` (basically updated `config_template.cfg` to mirror the changes made to `comfig.cfg` over the time, including the most recent ones).
- Added/updated both `ragdolls=hidden` and `ragdolls=low` as an alias to `ragdolls=hidden`. Not sure which of them should exist, but I have added both and marked them as deprecated.
  - In a future update, we can just remove both of them. Personally, I would prefer to remove both in this next release, since it's a big one after almost one year.
- Some misc. changes here and there.

Please let me know if new changes are needed. Thanks! I'll mark this PR as ready for review.

-----------------

One thing I noticed: since we have been changing all instances of "co**n**fig" to "co**m**fig", should we change "co**n**fig_template.cfg" to "co**m**fig_template.cfg" too?

Tiagoquix:
@mastercoms Btw, in regards to `r_radialfog_force`, should we set it to `0` in the Very Low preset since we're enabling it globally?

Tiagoquix:
@mastercoms updated to support latest changes to staging branch + updated initial post to better reflect current changes.

some things I want to reiterate:
1. ragdolls=low and ragdolls=hidden can be removed completely instead of just being deprecated. last comfig stable release was more than a year ago so I don't think it would be a problem to make users adjust their modules.
2. please take a look at my last comment regarding radial fog force.
3. also please take a deeper look into the mm_override.cfg changes (and the file in general) to see if we can remove more ConVars from there or if we should keep/add some or more of them.

many thanks for your attention!

mastercoms (approved):

Tiagoquix:
> @mastercoms Btw, in regards to `r_radialfog_force`, should we set it to `0` in the Very Low preset since we're enabling it globally?

friendly ping @mastercoms

PR #783, opened 2025-03-08
fix: wrong cvar name in FAQ

uen420:
The correct cvar is `r_drawmetalspark`. 

`r_drawimpactspark` does not exist in the game and is reflected with the console not returning autocomplete or a valid command execution, whereas it does for `r_drawmetalspark`.


mastercoms:
I think this was just a copy-paste error since the config uses the correct names (replaced only the last word for each in the list). Thanks for the catch.

mastercoms (approved):

PR #787, opened 2025-05-19
Changes to the ragdoll module

Tiagoquix:
- Removes `ragdolls=hidden` and `ragdolls=low` entirely.
  - It has been more than one year since the last stable release, so it's fine if we remove them. People can adjust themselves; no big deal.
- Renamed `ragdolls=medium` to `ragdolls=low`.
  - Consistency with other modules. Currently, it's `off`, `medium` and `high`. `low` and `high` without `medium`, in contrast to `medium` and `high` without `low`, makes more sense.
- Miscellaneous commands/comments adjustments.


mastercoms:
I don't agree with removing an option even if it's been a year since the last release, it doesn't change the user impact. We always give users a chance before breaking stuff.

Tiagoquix:
Considering the next release will be a big release, I think it's fine to pass these changes.

mastercoms:
Sorry, I just don't agree, a large release increases the load on people already, I don't think they should have to track every single change and have their config break, especially when it's something like ragdolls suddenly being enabled which is a huge performance impact for people

Tiagoquix:
Fine then. I will keep this PR open for the release after this next one, so we can keep this work for the future. Thanks.

PR #791, opened 2025-05-19
mm_override.cfg changes

Tiagoquix:
Added `mat_filterlightmaps` and `props_break_max_pieces`.

Removed `mat_supportflashlight` (the game automatically determines the ideal value if necessary) and a ragdoll-related ConVar since we don't have the others (cl_ragdoll etc.).


mastercoms:
Those are needed because mismatches for 5.0f are recorded, the shader switching for flashlight must be forced off,

Tiagoquix:
Why the shader switching for the flashlight must be forced off? That ConVar is set to `-1` by default, which determines the best value according to hardware.

mastercoms:
According to the game, not the hardware. At startup sets it to 0, since TF2 does not use flashlights.

PR #792, opened 2025-05-19
Obliterate config_template.cfg

Tiagoquix:
The template lacks some features of mastercomfig since some commands don't work in the `autoexec.cfg` alone.

It also creates an additional burden of mantainence since we have to edit both `comfig.cfg` and `config_template.cfg` when updating the comfig, along with changes that are only relevant for `config_template.cfg` (such as `(does not work in autoexec)` comments and uncomenting some ConVars/adding new lines that only impact the template).

It also tends to sometimes be forgotten because it isn't all that relevant and so lacks updates from the main `comfig.cfg` file.

I've personally updated the file recently to match most of `comfig.cfg` changes, but I think this file is not worth having because, _in addition to creating a burden of maintanence and some things not working with the template alone,_ it is also unsupported (since it is expected that the user will use the VPK release) and most of what the template does can be done using the `None` preset and applying the commands yourself.


mastercoms:
autoexec is still used by people

PR #793, opened 2025-05-23
Force radial fog OFF for the Very Low preset

Tiagoquix:
Possibly a minor performance improvement in comparison to having radial fog? (Not sure tbh.)


mastercoms:
I don't think it's a performance impact

PR #810, opened 2025-10-05
Make comfig echo be executed later

Tiagoquix:
"Hides" ugly messages such as "app/addons.cfg not present; not executing" appearing after the echo.

Runs after all commands, except dynamic_background module.


mastercoms:
Isn't actually a problem, this would clear selection messages and the actual problem is the addons.cfg not being generated

PR #812, opened 2025-10-05
Rename post_processing default level to valve

Tiagoquix:


mastercoms:
I don't think `valve` is a good setting name

Tiagoquix:
@mastercoms what about `original` or `vanilla`?

PR #813, opened 2025-10-10
Consolidate load order between app and overrides

Katzenwerfer:
Looking through the files, I noticed that some scripts in the `app` directory are executed after those in the `overrides` directory. If I'm not mistaken, the intended behaviour is to always run scripts from the former before the latter.

This pull request modifies the scripts I identified to match the description of the [9.100.0](https://github.com/mastercomfig/mastercomfig/releases/tag/9.100.0) patch notes.

> - Your own files remain in cfg/overrides/, but the app will generate an cfg/app/ folder for itself to generate its own files without replacing your files

I will be watching for any changes you request in case you want to merge this.

### DIFF
--- a/config/mastercomfig/cfg/comfig/comfig.cfg
+++ b/config/mastercomfig/cfg/comfig/comfig.cfg
@@ -1981,7 +1981,7 @@ alias modules_c"exec overrides/modules.cfg"
 
 alias run_modules"exec comfig/modules_run.cfg"
 alias module_levels"exec comfig/module_levels.cfg"
-alias apply_overrides"modules_c;run_modules;exec overrides/autoexec.cfg;exec app/autoexec.cfg"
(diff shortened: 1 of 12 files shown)


mastercoms:
So in the class config case, since these are used in a different context being an optional feature which the user explicitly sets on their own, it does override. In the future if the app does any automatic/transparent configuration I would just add a pre overrides app folder and execute another set of configs in that place. Also keep in mind the patch note there is talking literally about replacing files, had they both been in the overrides folder.

Also, be wary of using the class config aliases are those are encouraged to be set by the user and thus the app configs have no business being inside there, or else the user may forget. Not to mention that those are much older than the app configs so people may have alias overrides which don't include them and I don't see a benefit to giving user the need to add it.

In short, there's an exception to the rules in terms of a strict interpretation of it but as it stands, in practice, user preference is always maintained. I don't think this PR is needed at this time. Thanks for reporting this and taking the time to fix it though!

